PR #30, opened 2014-05-10
Improvement/x3 seek

vtnerd:
Slight modification to x3-seek implementation after K-ballo's feedback.


Kojoley:
Does anyone remember what this PR had been fixing? It introduced a regression #658. I cannot come to a reproducer from the changes. It also diverged X3 seek implementation from Qi.

PR #34, opened 2014-06-05
correct incorrect context_debug output

cppljevans:
The problem which this solves was discussed in post:

http://article.gmane.org/gmane.comp.parsers.spirit.general/26803

In summary, the problem occurs because the statement:

  traits::post_transform(attr, attr_);

swaps the arguments (at least when they're recursive variants)
instead of copying source(attr_) to target(attr).

Also, change argument to context_debug from made_attr to
attr_ because that makes it clearer what's happening since
attr_ is what is explicitly passed to parse_rhs.


Kojoley:
Was replaced by #38

PR #40, opened 2014-06-14
Fixing line_pos_iterator column numbering bug and get_current_line bug.

tomoki:
This is cleaned up version of https://github.com/boostorg/spirit/pull/39.

This pull request fixes following bugs.
- line_pos_iterator's column numbering bug.
- line_pos_iterator's get_current_line bug.

And add regression test for line_pos_iterator.

### DIFF
--- a/test/support/regression_line_pos_iterator.cpp
+++ b/test/support/regression_line_pos_iterator.cpp
@@ -1,0 +1,45 @@
+//  Copyright (c) 2014 Tomoki Imai
+//
+//  Distributed under the Boost Software License, Version 1.0. (See accompanying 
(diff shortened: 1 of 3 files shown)


test/support/regression_line_pos_iterator.cpp
```diff
@@ -0,0 +1,45 @@
...
+            if (line == 1) {
+                BOOST_TEST_EQ(current_line,"012");
+            } else if(line == 2) {
+                BOOST_TEST_EQ(current_line,"345");
+            } else {
+                BOOST_TEST_EQ(current_line,"678");
+            }
+        }
+    }
+    return 0;
```
octopus-prime: Should be `return boost::report_errors();` to report the errors.

octopus-prime:
See #305 and #304

PR #53, opened 2014-08-28
karma: support boost::recursive_wrapper

purpleKarrot:


Kojoley:
I tried my best to find a case where this applies but I did not. Karma should never get a `boost::recursive_wrapper` value because it lives only inside `boost::variant`. Moreover Karma extracts values from a variant variable with `boost::get` like this:

```cpp
#include <boost/variant.hpp>
#include <vector>

struct foo; // Forward declaration

typedef boost::variant<int, boost::recursive_wrapper<foo> > bar;

struct foo
{
    int i;
    std::vector<bar> v;
}; 

int main()
{
    bar v = { foo{ 0, {} } };
    boost::get<foo>(v);
}
```

The only case when `boost::recursive_wrapper` can be exposed to Karma if you use it outside of `boost::variant` for some unknown reason, but in that case you should deal with this by you own.

purpleKarrot:
I don't remember exactly, but I think I wanted to print the following type:

```
using int_tree_t = boost::make_recursive_variant<
      int
    , std::vector< boost::recursive_variant_ >
    >::type;
```
Writing a parser works, because Qi supports recursive_wrapper.
Writing a generator with Karma does nor work out of the box.

Kojoley:
1. The `boost::recursive_variant_` is a different thing.
2. Karma handles both of them out of the box. https://wandbox.org/permlink/dF8gsRRiyjg7n410 https://wandbox.org/permlink/7K52BuIipKFKFNMy
3. Reopen if you have an MWE.

PR #68, opened 2014-11-06
Provide a workaround for the missbehaving gcc compiler when using decltype inside a template function with an assignment.

teajay-fr:
I have found a workaround for the gcc issue which prevent x3 from being usable with gcc.

It seems the compiler handles the assignment operator specially and can't resolve the type inside the decltype() inside a template function.

I didn't go through all the examples, I just updated the calc1.cpp example.
Tell me what you think of this. If the solution I proposed is viable, I'll patch the rest of the samples.

### DIFF
--- a/example/x3/calc1.cpp
+++ b/example/x3/calc1.cpp
@@ -15,7 +15,7 @@
 //  [ JDG June 6, 2014 ]        spirit x3
 //
 ///////////////////////////////////////////////////////////////////////////////
-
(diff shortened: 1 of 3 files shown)


timangus:
Does this mean that spirit x3 still doesn't work with GCC? I'm using 5.4 and I get  'parser::parse_rule' declared as an 'inline' variable error...

PR #77, opened 2015-04-12
Suppress always false warning.

Flast:


Flast:
UPDATE: I dropped second patch because it had been outdated by other PR.

Kojoley:
Superseded by #243

PR #79, opened 2015-04-15
Added method to the position_policy to get the current line and column i...

darkennight:
...n the output.

By forwarding these methodes from the position_sink, someone can access the current line and column of the output_iterator when creating a new terminal.
This is implemented as announce in the mailing list by @daminetreg and myself.

### DIFF
--- a/include/boost/spirit/home/karma/detail/output_iterator.hpp
+++ b/include/boost/spirit/home/karma/detail/output_iterator.hpp
@@ -79,6 +79,18 @@ namespace boost { namespace spirit { namespace karma { namespace detail
             return track_position_data.get_count();
         }
 
+	// return the current line in the output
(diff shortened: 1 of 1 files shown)


djowel:
Why get_out_line and get_out_column instead of simply get_line and get_column?

darkennight:
The get_count function is forwared by get_out_count. I named these to function in the same way. Maybe the idea was, that when calling it on the sink inside the generate function of a terminal is to know, it is for the output and there is no overlapping to functions of a normal output iterator.
But I can change it to simply get_line and get_column. What do you mean?

djowel:
Yes, there is no "in" anyway in an "output" iterator.

darkennight:
Well, I removed "out" from the function names.
Thank you for your input

PR #84, opened 2015-04-29
fix unused parameter

tabe:
This avoids a warning like:
```
/some/path/include/boost/spirit/home/lex/lexer/lexertl/functor_data.hpp:171:42: warning: unused parameter ‘it’ [-Wunused-parameter]
             void set_end(Iterator const& it) {}
                                          ^
```


djowel:
please comment it out instead, like: /*iter*/. it's our convention.

PR #85, opened 2015-04-29
x3::variant needs check-for-self in its member templates

JamesWidman:
I ran into some compile-time errors when trying to copy an object of a type derived from x3::variant.

The root of the issue seems to be the SFINAE check on the members operator=; namely: when variant's move-assignment operator template is considered during overload resolution, it deduces 'A&' for template parameter T, and then reference-collapsing transforms 'variant& &&' into 'variant&', and thus the template ends up being a better choice than the implicitly-declared copy-assignment operator. This leads to a nice tall error message.

It seems like we just need a test for "self" as part of the SFINAE check on that operator template---which we used to have, but it appears to have been commented out at some point (for reasons not immediately clear to me).

In this commit, four of the member templates now use the same compact SFINAE check in the same way (as a default template argument).  It seems to do the right thing for me.


K-ballo:
I think you are targeting the wrong branch, I doubt this change would touch 123 files. PRs should target the `develop` branch.

JamesWidman:
Yes... I'm not sure how that happened; I really only touched one file, and there's only one commit.

(it's my first pull request ever; apologies.)

JamesWidman:
Ah: I see.  I initially had master checked out locally, then switched to develop.  So the next time I pushed, I guess github thought that I wanted to push my develop into master...?  But I really just wanted to push a new commit onto the existing remote "develop" without touching master at all.

teajay-fr:
You can do that with GitHub, you have to edit the target branch in 
GitHub when you configure your pull request.
I had that problem too a couple of times.

Regards,

Thomas Bernard

JamesWidman:
Cool; thanks!

I'll close this one and open a new one.

PR #86, opened 2015-04-29
Updated the check-for-self in x3::variant

JamesWidman:
I ran into some compile-time errors when trying to copy an object of a type derived from x3::variant.

The root of the issue seems to be the SFINAE check on the members operator=; namely: when variant's move-assignment operator template is considered during overload resolution, it deduces 'A&' for template parameter T, and then reference-collapsing transforms 'variant& &&' into 'variant&', and thus the template ends up being a better choice than the implicitly-declared copy-assignment operator. This leads to a nice tall error message.

It seems like we just need a test for "self" as part of the SFINAE check on that operator template---which we used to have, but it appears to have been commented out at some point (for reasons not immediately clear to me).

In this commit, four of the member templates now use the same compact SFINAE check in the same way (as a default template argument). It seems to do the right thing for me.

### DIFF
--- a/include/boost/spirit/home/x3/support/ast/variant.hpp
+++ b/include/boost/spirit/home/x3/support/ast/variant.hpp
@@ -14,6 +14,7 @@
 #include <boost/variant.hpp>
 #include <boost/mpl/list.hpp>
 #include <boost/type_traits/is_base_of.hpp>
+#include <type_traits>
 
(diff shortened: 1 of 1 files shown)


JamesWidman:
Well that looks much more sane than PR #85. (:

K-ballo:
> Well that looks much more sane than PR #18 

Indeed! :)

djowel:
nice!

PR #91, opened 2015-05-16
Rexpr fixes

ahmedcharles:


djowel:
Thanks for this fixes. Would you mind taking off the copyrights? I'd accept a joint copyright only if there's significant contribution that merits the status of co-authorship. I don't think it is appropriate for simple fixes at this point. Thanks!

djowel:
Thanks for the tweaks!

PR #93, opened 2015-05-16
Allow rule.hpp to be included independently.

ahmedcharles:
Have rexpr include rule.hpp independently as a check.

### DIFF
--- a/example/x3/rexpr/rexpr_full/rexpr/rexpr.hpp
+++ b/example/x3/rexpr/rexpr_full/rexpr/rexpr.hpp
@@ -7,8 +7,9 @@
 #if !defined(BOOST_SPIRIT_X3_REXPR_HPP)
 #define BOOST_SPIRIT_X3_REXPR_HPP
 
-#include <boost/spirit/home/x3.hpp>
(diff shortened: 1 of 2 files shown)


djowel:
In X3, my inclination is to simplify the include with only one include: #include <boost/spirit/home/x3.hpp>. Do you find that the situation causes a real slowdown? If so, I'd like to see some numbers on the compilation speed of both independent and single include. Most users will include most of X3 anyway.

ahmedcharles:
That makes sense and I don't have empirical evidence that suggests that it isn't viable. Though, I imagine that eventually, someone is going to want to only #include <boost/spirit/home/x3/nonterminal/rule.hpp> rather than all of X3, especially since one of the goals of having the rule separation is improved compile times.

And I'm a big fan of every header including it's dependencies.

djowel:
Agreed. Perhaps I can provide docs for those who want independent includes. For the tutorials though, I'd go with a single include for ease of use.

ahmedcharles:
I understand the objection now. This change now still uses the general include for the example, but allows rule.hpp to be included while also compiling.

PR #96, opened 2015-05-18
Make annotation.hpp generic and move to support/utility/annotate_on_success.hpp.

mlang:
Since X3 promotes x3::variant, we can actually make annotation.hpp generic,
avoiding copy-pastism in all the projects that make use of basic error
reporting.


djowel:
Wonderful! I wanted to do that soon. You beat me to it. Thanks, Mario! You rock!

djowel:
Next step would be error handler :-)

PR #102, opened 2015-05-21
x3::matches.

mlang:


teajay-fr:
Would it not be better to put this directive in the extensions directory ?

djowel:
wonderful! thanks!

djowel:
yes, this should be in directives.

djowel:
- i moved it to directives
- i also removed extensions. no need to confuse users

PR #103, opened 2015-05-22
Fix include paths.

mlang:
Now the x3 testsuite builds again.


djowel:
ooops. sorry about that.

PR #104, opened 2015-05-23
Add support for the binary parsers

teajay-fr:
This is not polished yet, and doesn't support floating point types. Consider this as pre-review pull request.

Comments are expected and welcome.

### DIFF
--- a/include/boost/spirit/home/x3/binary.hpp
+++ b/include/boost/spirit/home/x3/binary.hpp
@@ -1,0 +1,16 @@
+/*=============================================================================
+    Copyright (c) 2001-2011 Hartmut Kaiser
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+==============================================================================*/
+#if !defined(BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0906AM)
+#define BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0906AM
+
+#if defined(_MSC_VER)
+#pragma once
+#endif
+
+#include <boost/spirit/home/x3/binary/binary.hpp>
+
+#endif
--- a/include/boost/spirit/home/x3/binary/binary.hpp
+++ b/include/boost/spirit/home/x3/binary/binary.hpp
@@ -1,0 +1,209 @@
+/*=============================================================================
+    Copyright (c) 2001-2011 Hartmut Kaiser
+    Copyright (c) 2001-2011 Joel de Guzman
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+==============================================================================*/
+#if !defined(BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0808AM)
+#define BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0808AM
+
+#if defined(_MSC_VER)
+#pragma once
+#endif
+
+#include <boost/spirit/home/x3/core/parser.hpp>
+#include <boost/spirit/home/x3/core/skip_over.hpp>
+#include <boost/spirit/home/x3/support/traits/move_to.hpp>
+#include <cstdint>
+
+
+#include <boost/endian/conversion.hpp>
+#include <boost/endian/arithmetic.hpp>
+#include <boost/fusion/include/at.hpp>
+#include <boost/mpl/or.hpp>
+#include <boost/type_traits/is_integral.hpp>
+#include <boost/type_traits/is_enum.hpp>
+#include <boost/type_traits/is_floating_point.hpp>
+#include <boost/config.hpp>
+
+namespace boost { namespace spirit { namespace x3
+{
+#if 0
+    ///////////////////////////////////////////////////////////////////////
+    template <typename Encoding, typename Attribute>
+    struct get_info<literal_char<Encoding, Attribute>>
+    {
+        typedef std::string result_type;
+        std::string operator()(literal_char<Encoding, Attribute> const& p) const
+        {
+            return '\'' + to_utf8(Encoding::toucs4(p.ch)) + '\'';
+        }
+    };
+    template <BOOST_SCOPED_ENUM(boost::endian::endianness) bits>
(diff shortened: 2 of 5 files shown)


include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+                *bytes++ = *it++;
+            }
+
+            first = it;
+            x3::traits::move_to(endian::conditional_reverse< endian, endian::order::native>(attr_), attr_param);
+            return true;
+        }
+
+        template <typename V>
+        binary_lit_parser< V, T, endian, bits> operator()(V n) const
```
vtnerd: Should the template be removed from this function? If it were limited to type `T`, a user would get an error immediately if it couldn't be converted. Currently a conversion error won't occur until the parse function of `binary_lit_parser`.

include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+#endif
+            attr_ = n;
+#if defined(BOOST_MSVC)
+#pragma warning(pop)
+#endif
+
+            unsigned char* bytes = reinterpret_cast<unsigned char*>(&attr_);
+
+            Iterator it = first;
+            for (unsigned int i = 0; i < sizeof(attr_); ++i)
```
vtnerd: Should this be done in the constructor? The user would pay a small penalty if the parse function was never called, but if it were called more than once, it would be a small benefit.

Similarly, the move_to function below does any endianness conversion lazily. Its possible to do the conversion immediately in the constructor too. However, a native type could be storing a value in the wrong endianness (which goes against the purpose of the `boost::endian` library).

If the v2 version does it this way, perhaps it should duplicated.

include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+            {
+                return "big-endian binary";
+            }
+        };
+    }
+#endif
+    ///////////////////////////////////////////////////////////////////////////
+    template <typename V, typename T
+      , BOOST_SCOPED_ENUM(boost::endian::order) endian, std::size_t bits>
+    struct binary_lit_parser
```
djowel: Maybe use enum class here instead?

include/boost/spirit/home/x3/binary.hpp
```diff
@@ -0,0 +1,16 @@
...
+    Copyright (c) 2001-2011 Hartmut Kaiser
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+==============================================================================*/
+#if !defined(BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0906AM)
+#define BOOST_SPIRIT_X3_BINARY_MAY_08_2007_0906AM
+
+#if defined(_MSC_VER)
+#pragma once
```
djowel: This pragma once for msvc has ben bugging me forever :) 1) is it really that significant? 2) are there no pragma once in other compilers. this is quite prevalent in spirit.

include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+            // Even if the endian types are not pod's (at least not in the
+            // definition of C++03) it seems to be safe to assume they are
+            // (but in C++0x the endian types _are_ PODs).
+            // This allows us to treat them as a sequence of consecutive bytes.
+            boost::endian::endian_arithmetic<endian, T, bits> attr_;
+
+#if defined(BOOST_MSVC)
+// warning C4244: 'argument' : conversion from 'const int' to 'foo', possible loss of data
+#pragma warning(push)
+#pragma warning(disable: 4244)
```
djowel: remove msvc gunk for now?

include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+#endif
+            attr_ = n;
+#if defined(BOOST_MSVC)
+#pragma warning(pop)
+#endif
+
+            unsigned char* bytes = reinterpret_cast<unsigned char*>(&attr_);
```
djowel: use auto ;-)

include/boost/spirit/home/x3/binary/binary.hpp
```diff
@@ -0,0 +1,209 @@
...
+        {
+            x3::skip_over(first, last, context);
+
+            attribute_type attr_;
+            unsigned char* bytes = reinterpret_cast<unsigned char*>(&attr_);
+
+            Iterator it = first;
+            for (unsigned int i = 0; i < sizeof(attr_); ++i)
+            {
+                std::cout <<"Hey"<<std::endl;
```
djowel: :-)

include/boost/spirit/home/x3/binary/binary.hpp
djowel: This might undergo some changes still once I get 'lazy' parsers back in. But you can ignore my comment form now if that does not seem make any sense.

include/boost/spirit/home/x3/binary.hpp
K-ballo: 1) Not that significant, as compilers have learned to recognize header guards to introduce the same behavior `#pragma once` would.
2) It's supported by several compilers according to the definition of `BOOST_HAS_PRAGMA_ONCE`.

teajay-fr: According to Wikipedia most of the c++ compiler support it, except solaris studio. I'll check how BOOST_HAS_PRAGMA_ONCE is supposed to be used.

include/boost/spirit/home/x3/binary/binary.hpp
teajay-fr: Yes. Doesn't make sense right now.

include/boost/spirit/home/x3/binary.hpp
djowel: I think we should just remove all pragma once. Header guards should suffice.

teajay-fr:
Just pushed a finalized version of the binary parser. Floating point support will have to wait until the endian library support it. It is planned for 1.59.

djowel:
I'll merge this. let us refine later.

PR #106, opened 2015-05-25
X3: Add support for custom character filters on the symbols parser

teajay-fr:
Just added an additional template parameter to the symbols parser as in qi, except it's not the filter directly, only a helper struct to get hold of the correct filter. This was necessary to continue to support case insensitive parsing.

### DIFF
--- a/include/boost/spirit/home/x3/string/symbols.hpp
+++ b/include/boost/spirit/home/x3/string/symbols.hpp
@@ -36,11 +36,22 @@
 
 namespace boost { namespace spirit { namespace x3
 {
+    template < typename Encoding >
+    struct get_default_symbol_filter
+    {
+        template <typename Context>
+        static auto get(Context const& context) -> decltype(get_case_compare<Encoding>(context))
+        {
+            return {};
+        }
+    };
+
     template <
         typename Encoding
       , typename T = unused_type
+      , typename GetFilter = get_default_symbol_filter<Encoding>
       , typename Lookup = tst<typename Encoding::char_type, T> >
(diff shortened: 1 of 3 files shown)


include/boost/spirit/home/x3/string/symbols.hpp
```diff
@@ -36,11 +36,22 @@
...
+    template < typename Encoding >
+    struct get_default_symbol_filter
+    {
+        template <typename Context>
+        static auto get(Context const& context) -> decltype(get_case_compare<Encoding>(context))
+        {
+            return {};
+        }
+    };
+
```
saki7, line 48: I'm afraid that this approach will lead to ODR violation. The code itself seems okay, but we must think of more complex situations.

A user would expect that they're the only person who is specializing `get_default_symbol_filter` for their arbitrary `Encoding`, but in reality, if any conflicting definitions (even on 3rd party libraries) exist in different translation units, it will trigger UB.

That's very likely to happen, since the predefined encodings such as `char_encoding::unicode` would normally be passed to the `symbols_parser`, in contrast to the custom type `braille::encoding::dots` used in your example.

Proposed code must be at least not using the word `default`, but even so, we definitely need a robust customization point which correctly handles potentially conflicting definitions.

It shall be implemented in a more C++20-ish way, such as explicitly requiring a CPO with specific signatures that is defined per each parser primitives.

That said, I don't think (and I can't demand to) implement that kind of huge change on top of this PR.

saki7: Well, a similar seems to be implemented already in 2025. But that's a different story, I guess.
Apologies that we couldn't review your PR in 2015.

PR #118, opened 2015-06-15
X3: Make use of Fusion member type deduction to simplify examples.

mlang:


daminetreg:
Thanks @mlang :smile:

PR #123, opened 2015-06-18
X3: Add another missing const&.

mlang:
A tiny change which results in 7kb less binary size and 17.8s to 17.2s compile
time reduction (GCC).

### DIFF
--- a/include/boost/spirit/home/x3/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/rule.hpp
@@ -53,7 +53,7 @@ namespace boost { namespace spirit { namespace x3
         static bool const force_attribute =
             force_attribute_;
 
-        rule_definition(RHS rhs, char const* name)
+        rule_definition(RHS const& rhs, char const* name)
           : rhs(rhs), name(name) {}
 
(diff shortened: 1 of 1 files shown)


vtnerd:
Should X3 change its constructors to the "[pass by value then move idiom](http://stackoverflow.com/questions/21035417/is-the-pass-by-value-and-then-move-construct-a-bad-idiom)"? The operator overloads would have to be updated too. However, this could [increase the size of the binary](https://web.archive.org/web/20140205194657/http://cpp-next.com/archive/2009/08/want-speed-pass-by-value/) in some situations - does anybody know how much of a difference moves could make for expression templates like this? I think there are only a few cases where dynamic memory is held anyway.

djowel:
I made it pass by value initially. It seems, as mlang reported, that doing o increases compile time and binary size. You can check it out by reverting some of mlang's previous commits.

mlang:
I think the "pass by value and move" idiom is useful when the signature
has several arguments which demand different ref-ness treatment.  If you
just pass a single arg, you might as well pass it by const& to avoid copies
and moves.  As mentioned in this thread, moving a spirit parser is
basically just a copy anyway,
since they dont contain anything movable (typically).

PR #126, opened 2015-06-21
attribute type specification used `basic_string`

sehe:
It was [brought to my attention](http://stackoverflow.com/questions/11421430/string-parser-with-boost-variant-recursive-wrapper/11422737?noredirect=1#comment49970878_11422737) on SO that the documentation seems to be (confusingly) in error.

I admit I get more questions about "why can't I assign the synthesized attribute to a `std::string`" than I could explain. So I guess this should be fixed.

### DIFF
--- a/doc/qi/string.qbk
+++ b/doc/qi/string.qbk
@@ -87,7 +87,7 @@ not defined in __primitive_parser_concept__.
     [[Expression]       [Attribute]]
     [[`s`]              [__unused__]]
     [[`lit(s)`]         [__unused__]]
-    [[`ns::string(s)`]  [`std::basic_string<T>` where `T`
+    [[`ns::string(s)`]  [`std::vector<T>` where `T`
                         is the underlying character type
                         of `s`.]]
 ]


djowel:
The synthesized attribute of string parser is really a std::string (see boost/spirit/home/qi/lit.hpp): 

        typedef std::basic_string<char_type> string_type;
   ...
       template <typename Context, typename Iterator>
        struct attribute
        {
            typedef typename mpl::if_c<
                no_attribute, unused_type, string_type>::type
            type;
        };

What am I missing?

sehe:
Well. That's embarassing. You're right of course.

Then again, I'm not sure what caused the problem in that SO question I linked to anymore. I'll need to revisit that (or maybe you can give it a quick scan). It's rather old, so maybe you recognize an already fixed issue in there :/

Other than that, disregard this pull request :)

djowel:
Yes. If you can distill the problem further, then I can go take a look at it and see what's going on.

djowel:
I had this as 'wontfix' for now, but let's continue with the investigation.

Kojoley:
> The synthesized attribute of string parser is really a std::string

Actually `qi::assign_to`(`x3::move_to`) receives a pair of iterators and not the `std::string`. I have opened #321.

PR #132, opened 2015-07-05
return type via decltype(auto)

bebuch:


bebuch:
It's a nice simplification from C++14 ;-)

http://en.cppreference.com/w/cpp/language/auto

PR #141, opened 2015-07-14
Unitialized data fix

vtnerd:
This is a fix for the issue posted on the mailing list. I don't see any other unitialized data values. I don't think a test can be written unless `n` is initialized, otherwise the test will be unreliable (n _usually_ equals some small nearly zero value on my box, but it is undefined ....).


vtnerd:
Also note, this is issued against the developer branch, and not master ...

djowel:
I'll merge this one and merge to master asap. 

I understand now why the test might be unreliable. I guess there's no easy way to deal with this other than test against valgrind.

PR #142, opened 2015-07-21
Fixed modifier bug with karma::left_align

vtnerd:
I found a bug in the factory function for `karma::left_align`. `upper[left_align(10, 'a')['x']]` currently fails to compile because the constructor for the `left_align` object is given a different type. I applied the modifiers to the pad character because `right_align` does, and that appears to be the goal. However, should it be a direct change? In other words should `upper[left_align(10, 'a')['x']]` output `Xaaaaaaaaa` or `XAAAAAAAAA`?

### DIFF
--- a/include/boost/spirit/home/karma/directive/left_alignment.hpp
+++ b/include/boost/spirit/home/karma/directive/left_alignment.hpp
@@ -275,10 +275,10 @@ namespace boost { namespace spirit { namespace karma
 
         template <typename Terminal>
         result_type operator()(Terminal const& term, Subject const& subject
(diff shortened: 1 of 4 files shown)


djowel:
Do we want this in master? If so, please do a PR to develop first, let the tests cycle and I'll do the merge to master.

hkaiser:
LGTM, this should output `XAAAAAAAAAA`. But it has to go through develop.

vtnerd:
Sorry about that, the target branch cannot be modified, so I will open a new request. I also added the same tests for `center`, which passed.

PR #144, opened 2015-07-31
Fixed bug in karma::duplicate attribute reporting

vtnerd:
An expression such as `generate(..., (duplicate[double_ << double_] | "foo"), 2.0)` will output "foo" even though `duplicate[double_ << double_]` should never fail. The problem is that `duplicate` adds a reference to the reported attribute type. The `is_convertible` check will fail, causing the first branch of the alternative to be ignored at compile-time.

All of the tests were run on OSX with this patch - no issues. I leave it to the spirit braintrust to determine whether this could cause any issues with code in the wild.


hkaiser:
LGTM, thanks!

vtnerd:
A gentle bump on this pull request ...

PR #145, opened 2015-08-05
Fixing race condition on static init

biocomp:
Static variables depend on each other and when static init is not thread-safe, this leads to race condition and failures (on, say, Visual Studio 2013 update 4).


biocomp:
Closed pull request to main because I was asked to create one for develop

PR #146, opened 2015-08-05
Fixing race condition on static init

biocomp:
Static variables depend on each other and when static init is not thread-safe, this leads to race condition and failures. Ticket: https://svn.boost.org/trac/boost/ticket/11514

### DIFF
--- a/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
+++ b/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
@@ -135,8 +135,9 @@ namespace boost { namespace spirit { namespace qi { namespace detail
         inline static bool add(T& n, Char ch, mpl::true_) // checked add
         {
             // Ensure n *= Radix will not overflow
+            // Not reusing max static variable to avoid race condition on old compilers
             static T const max = (std::numeric_limits<T>::max)();
-            static T const val = max / Radix;
+            static T const val = (std::numeric_limits<T>::max) / Radix;
 
             if (n > val)
                 return false;
@@ -167,8 +168,9 @@ namespace boost { namespace spirit { namespace qi { namespace detail
         inline static bool add(T& n, Char ch, mpl::true_) // checked subtract
         {
             // Ensure n *= Radix will not underflow
+            // Not reusing min static variable to avoid race condition on old compilers
             static T const min = (std::numeric_limits<T>::min)();
-            static T const val = (min + 1) / T(Radix);
+            static T const val = ((std::numeric_limits<T>::min) + 1) / T(Radix);
             if (n < val)
                 return false;
 
--- a/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
+++ b/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
@@ -31,6 +31,7 @@
 #include <boost/mpl/bool.hpp>
 #include <boost/mpl/and.hpp>
 #include <boost/limits.hpp>
+#include <boost/integer_traits.hpp>
 
 #if defined(BOOST_MSVC)
 # pragma warning(push)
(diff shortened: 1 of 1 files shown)


K-ballo:
The fix isn't actually a fix, is it? I'm surprised that you see a tear because of lack of magic statics on a platform that VS supports, but if you do see it I don't see how this fixes it. Would you explain what the issue is and how to reproduce it?

biocomp:
The repro is just starting multiple threads which trigger the parser method. And though rarely, race condition happens and val becomes 0 (for me). I don't have a nice repro, I was testing with my ugly actual code.

I guess you're right saying that this is not complete fix. Indeed, there's still a problem with static init, it's still not thread safe. BTW, good link about VS's implemenhtation of the thing: http://www.lenholgate.com/blog/2014/04/bitten-by-scoped-static-initialisation-in-c---turn-on-visual-studios-c4640-warning.html

Basically, the static init is this:
1. Check if initialized.
2. If yes, skip init.
3. Init.

Clearly, two threads may check flag and go init variable several times, or (worse), thread may see "initialized" being set and skip init, while other thread did not actually finish the init. That second scenario will give problems with my fix as well.

However, using TWO variables together makes things worse. If they're independent, init will happen EVENTUALLY, so maybe some threads will get incorrect value on the first run, but some time later one of the threads will set static variables and from there, it'll have valid value.

With two variables, one scenario is:
1. Race condition happens, and our current thread sees "max" being initialized, while the other thread has not done that yet, so max has some garbage value.
2. Our thread proceeds initializing "val" with incorrect max value, and stores "val" to be incorrect forever (well, for the live of the process).

That is what I've been experiencing, and to me, the fix is good enough.

Now, for C++11 (with constexpr support, not msvc) compilers, of course constexpr should be used. But I'm not sure what boost magic to use (and if there's any) to fix this for both constexpr supporting compilers and for older ones.

djowel:
> Now, for C++11 (with constexpr support, not msvc) compilers, of course constexpr 
> should be used. But I'm not sure what boost magic to use (and if there's any) to fix 
> this for both constexpr supporting compilers and for older ones.

BOOST_CONSTEXPR_OR_CONST perhaps?

biocomp:
> BOOST_CONSTEXPR_OR_CONST perhaps?
This will make it to be const for non-constexpr compilers. Depending on compiler, it could be all optimized-out, but in worst-case scenario, call to max() will be there and done every time. Could affect performance. Not sure if that is really a big problem.

djowel:
I think a better approach is to use boost integer_traits::const_min and const_max, if T is a built-in integer, at least for compilers without constexpr.

K-ballo:
If the operations are expected to be cheap, using automatic storage instead of static storage would be a fix for the potential race. I'm having trouble imagining a bounded integer for which those operations or the values themselves would be costly.

If the operations are expected or suspected to be expensive, making them global statics would solve the potential race for compilers without magic statics. Implementations with magic statics would potentially benefit as well, magic statics are not free as they involve some sort of `call_once`.

djowel:
in most cases, T is a builtin integer or a builtin floating point type. i think making them use const with automatic storage is the best approach plus boost integer_traits::const_min and const_max when constexpr is not available.

djowel:
Also, IIRC, this code won't be hit when T is a floating point type because check_overflow will be false for these types.

biocomp:
I guess I can't get away with my simple change, can't I? :) But you're right of course, let me see what would be the nicest way to use some sort of compile-time constants.

Also, with all that modular movement, is it OK to take dependency on integer traits library (if not already depending on it).

biocomp:
Updated the code to use compile-time constants.

djowel:
Wonderful! Now it remains to be seen if the tests will fail on float or double T. I assume you ran the tests? Thanks a lot!

djowel:
integer traits is fairly low level and I assume it's part of the core boost. in any case, i don't mind the extra dependency on such a trivial header.

biocomp:
I did run the tests, they're passing. I only tested msvc-12 and msvc-14 on windows though.

The description of the class I'm touching says it's only for integers, so there should be no concern about floats:

    //  positive_accumulator/negative_accumulator: Accumulator policies for
    //  extracting integers. Use positive_accumulator if number is positive.
    //  Use negative_accumulator if number is negative.

djowel:
Yes, there shouldn't be a concern.

PR #148, opened 2015-08-12
fixed 1.59.0 rc3: compiler warnings in real_impl.hpp

tobias-loew:
explicit type conversions & static_cast to avoid compiler
warning
warning C4244: '=' : conversion from 'unsigned __int64' to
'double',
possible loss of data
in real_impl.hpp


djowel:
Thanks!

tobias-loew:
I made a PR against develop but there the type_cast in line 72 is not included since both have type int. In master max_exp has type size_t which will result in a warning

Kojoley:
Was replaced by #149

PR #149, opened 2015-08-13
fixed 1.59.0 rc3: compiler warnings in real_impl.hpp

tobias-loew:
explicit type conversions & static_cast to avoid compiler
warning
warning C4244: '=' : conversion from 'unsigned __int64' to
'double',
possible loss of data
in real_impl.hpp


tobias-loew:
type_cast in line 72 is not included since both have type int. In master-branch max_exp has type size_t which will result in a compiler-warning

PR #150, opened 2015-08-14
prevent unused warnings in x3

bebuch:
I'm not quite sure about the macro magic in "include/boost/spirit/home/x3/nonterminal/rule.hpp" :/

### DIFF
--- a/include/boost/spirit/home/x3/auxiliary/attr.hpp
+++ b/include/boost/spirit/home/x3/auxiliary/attr.hpp
@@ -40,8 +40,8 @@ namespace boost { namespace spirit { namespace x3
 
         template <typename Iterator, typename Context
           , typename RuleContext, typename Attribute>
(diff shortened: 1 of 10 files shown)


tomilov:
Defenitely bad idea.

bebuch:
Do you mean only the macro magic change or the pull request as a whole? I looked at it more closely and think it is okay, but it is the first time I tried to understand boost.preprocessor.

It would be nice of you if you explain why it is a bad idea. :-)

djowel:
For unused arguments, the preferred style is to comment out the argument name instead of deleting. Example: 

bool parse(Iterator& /*first*/, Iterator const& /*last*/
-          , Context const& /*context*/, RuleContext&, Attribute& attr_) const

bebuch:
Okay, I will do that, thank you!

bebuch:
Close pull request. I will re-pull with simpler changes.

PR #153, opened 2015-08-31
reenabled support for BOOST_FUSION_ADAPT_ADT adapted structs

tobias-loew:
since commit 69cc5370381e48352dc01c708599e9b49b7460f0 (hkaiser hkaiser
authored on 7 Nov 2009) spirit support BOOST_FUSION_ADAPT_ADT adapted
structs was broken

### DIFF
--- a/include/boost/spirit/home/qi/detail/assign_to.hpp
+++ b/include/boost/spirit/home/qi/detail/assign_to.hpp
@@ -18,6 +18,7 @@
 #include <boost/spirit/home/qi/detail/attributes.hpp>
 #include <boost/spirit/home/support/container.hpp>
 #include <boost/fusion/include/copy.hpp>
+#include <boost/fusion/adapted/struct/detail/extension.hpp>
 #include <boost/ref.hpp>
 #include <boost/range/iterator_range.hpp>
 
@@ -200,6 +201,16 @@ namespace boost { namespace spirit { namespace traits
         call(boost::optional<T> const& val, Attribute& attr)
         {
             assign_to(val.get(), attr);
+        }
+    };
+
+    template <typename Attribute, int N, bool Const, typename T>
+    struct assign_to_attribute_from_value<fusion::extension::adt_attribute_proxy<Attribute, N, Const>, T>
+    {
+        static void
+        call(typename T const& val, typename fusion::extension::adt_attribute_proxy<Attribute, N, Const>& attr)
+        {
+            attr = val;
         }
     };
 


djowel:
Great you found a fix!

tobias-loew:
No problem - I'm happy to help such a great project.

sehe:
It seems there are still some loose ends with this. While answering http://stackoverflow.com/a/33468657/85371 I found that ADAPT_ADT is till practically broken on 1.59/develop with the permutation parser.

In the below testcase

```c++
        ("a=" >> qi::char_ >> ';')
    >> ("b=" >> qi::char_ >> ';')
```

compiles, but

```c++
("a=" >> qi::char_ >> ';')
^ ("b=" >> qi::char_ >> ';')
```

does not.

Full sample: **<kbd>[Live On Coliru](http://coliru.stacked-crooked.com/a/e1324cad918fc3d7)</kbd>**

```c++
#include <boost/fusion/adapted.hpp>
#include <iostream>

class Foo {
    public:
    char const& getA() const { return a; } 
    char const& getB() const { return b; } 

    void setA(char value)    { a = value; } 
    void setB(char value)    { b = value; } 
    private:
    char a, b;
};

BOOST_FUSION_ADAPT_ADT(
        Foo,
        (char const&, char const&, obj.getA(), obj.setA(val))
        (char const&, char const&, obj.getB(), obj.setB(val))
    )


int main() {

    boost::spirit::istream_iterator f(std::cin), l; // input e.g. "a=a;b=b;"
    Foo foo;

    bool r = qi::parse(f, l,
#if 1 // FAILS TO COMPILE
                        ("a=" >> qi::char_ >> ';')
                    ^ ("b=" >> qi::char_ >> ';')
#else // COMPILES WITH FIX FROM PR #153
                        ("a=" >> qi::char_ >> ';')
                    >> ("b=" >> qi::char_ >> ';')
#endif 
                , foo);

    if (r)
        std::cout << "Parsed: " << boost::fusion::as_vector(foo) << "\n";
    else
        std::cerr << "Parse failed\n";
}
```

daminetreg:
I think the problem of ADAPT_ADT is that the proxy object it uses for each member is not transparent enough, I've this on my list of possible improvement in fusion, because I believe if we could make adapted ADTs equal to adapted structs  we could remove the code specific for ADAPT_ADT in spirit and there wouldn't be any differences anymore.

sehe:
Sounds like a plan. For the short term, it's broken though. 

Do have a look at Niebler's articles on "perfect proxies" or similar. He seems to have really thought about the dos and don'ts as well as new possibilities in c++14

daminetreg:
Thanks for the tips, Eric Niebler is awesome :smile: , but as you say it's no short term solution.

Kojoley:
@sehe the bug should be fixed in #376.

Superlokkus:
@Kojoley Seems it's not fixed, see my other example https://stackoverflow.com/questions/52225089/boost-fusion-adapt-adt-explodes-when-using-with-boostkarma . I used boost 1.69 (todays master branch)

Kojoley:
Hm, const proxy does not have assignment operator, so I am not sure what this PR was fixing then.

tobias-loew:
Hi, the reason for the initial PR was that adapted structs like:

    BOOST_FUSION_ADAPT_ADT(
    boost::uuids::uuid,
    (obj.data[0],  obj.data[0]  = val )
    (obj.data[1],  obj.data[1]  = val )
    (obj.data[2],  obj.data[2]  = val )
    (obj.data[3],  obj.data[3]  = val )
    (obj.data[4],  obj.data[4]  = val )
    (obj.data[5],  obj.data[5]  = val )
    (obj.data[6],  obj.data[6]  = val )
    (obj.data[7],  obj.data[7]  = val )
    (obj.data[8],  obj.data[8]  = val )
    (obj.data[9],  obj.data[9]  = val )
    (obj.data[10], obj.data[10] = val )
    (obj.data[11], obj.data[11] = val )
    (obj.data[12], obj.data[12] = val )
    (obj.data[13], obj.data[13] = val )
    (obj.data[14], obj.data[14] = val )
    (obj.data[15], obj.data[15] = val )
    )

could not be used in spirit as they result in compile errors in assign_to.hpp line 153 (version 1.68 with my changes from this PR commented out), when using the following code for parsing a uuid inside a grammar:

        qi::uint_parser<uint8_t, 16, 2, 2> hx;
        uuid %= hx >> hx >> hx >> hx >> lit('-') >> hx >> hx >> lit('-') >> hx >> hx >> lit('-') >> hx >> hx >> lit('-') >> hx >> hx >> hx >> hx >> hx >> hx;
        qi::rule<Iterator, boost::uuids::uuid(), skipper<Iterator>> uuid;

Tobias

tobias-loew:
@sehe I tried your example http://coliru.stacked-crooked.com/a/e1324cad918fc3d7 on VS 2015 (Update 3) and VS 2017 (Update 8.2) with Boost 1.68 and both version (sequence and permutation parser) compile and work as expected.

Kojoley:
@Superlokkus yes, see #396.

Superlokkus:
Thank you very much guys/gals, keep up the good work! (Hope I will be able to contribute somewhere in the near future)

PR #163, opened 2015-11-03
Facilitate the requirements to custom floating-point type (comparison)

tomilov:
The change allows the user to make custom floating-point type equality comparable only. operator == (double) is not required now.

### DIFF
--- a/include/boost/spirit/home/x3/support/numeric_utils/detail/extract_int.hpp
+++ b/include/boost/spirit/home/x3/support/numeric_utils/detail/extract_int.hpp
@@ -416,7 +416,7 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
                 {
                     if (count == 0) // must have at least one digit
                         return false;
-                    attr = 0;
+                    attr = Attribute(0);
                     first = it;
                     return true;
                 }
--- a/include/boost/spirit/home/x3/support/numeric_utils/extract_real.hpp
+++ b/include/boost/spirit/home/x3/support/numeric_utils/extract_real.hpp
@@ -142,7 +142,7 @@ namespace boost { namespace spirit { namespace x3
             bool neg = p.parse_sign(first, last);
 
             // Now attempt to parse an integer
-            T n = 0;
+            T n(0);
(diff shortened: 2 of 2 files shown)


tomilov:
Documentation states, that `T` should be constructible from `int` only (`T(int)`). Whether it should be possible in explicit or implicit way is unspecified, but provided expression imply, that `T` should be explicitly constructible from `int`.

vtnerd:
I don't see where the X3 documentation lists the requirements for `T`. The v2 documentation also states this:

| boost::spirit::traits::is_equal_to_one(n)  | Return true if n is equal to 1.0. Default implementation is provided for float, double and long double. |
  -------------------------------------------------------|---------------------------------------

which would have to be updated to reflect the changes you have just made since its testing for `1`, and not `1.0`. And `T` would still need a `operator==` for itself in the default implementation of `is_equal_to_one`, which isn't listed in the documentation.

tomilov:
I stick on [Spirit.V2 documentation](http://www.boost.org/libs/spirit/doc/html/spirit/qi/reference/numeric/real.html#spirit.qi.reference.numeric.real.minimum_expression_requirements_for__code__phrase_role__identifier__t__phrase___code_), and concentrate attention to following line of the table of `T`'s requirements:

**Expression** | **Semantics**
------------ | -------------
`T(0)` | Construct from an int.

djowel:
I think this makes sense. Unless there are more objections, I'm willing to merge this. The documentation should be augmented with this info. And, better yet, we need tests for such custom floating-point types.

tomilov:
If you decide to start writing tests from a scratch, then look at [this](https://github.com/tomilov/insituc/blob/master/include/floating_point_type_wrapper.hpp).

djowel:
What's your interest with this, BTW? Are you intending to use custom numerics with Spirit?

tomilov:
I intent to provide an example of using a wrapper of CGAL's algebraic numbers or mpfr/gmp for [my JIT-compiler project](https://github.com/tomilov/insituc) (for AST transformations and interpreter parts).

Kojoley:
Closed and opened to kick the CI.

PR #168, opened 2016-01-23
[MSVC] fix error C2995

hia3:
MSVC-2015 can't handle decltype in function template argument list.
C2995: function template has already been defined. Moving decltype out
of the function fixes the problem.

### DIFF
--- a/test/x3/actions.cpp
+++ b/test/x3/actions.cpp
@@ -4,6 +4,7 @@
     Distributed under the Boost Software License, Version 1.0. (See accompanying
     file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 =============================================================================*/
+#include <boost/bind.hpp>
 #include <boost/detail/lightweight_test.hpp>
 #include <boost/spirit/home/x3.hpp>
 #include <cstring>
@@ -22,6 +23,7 @@ auto fun1 =
 
 struct fun_action
 {
+    using result_type = void;
     template <typename Context>
     void operator()(Context const& ctx) const
     {
@@ -65,9 +67,9 @@ int main()
     }
 
     {
-        using namespace std::placeholders;
+        using namespace boost::placeholders;
         char const *s1 = "{42}", *e1 = s1 + std::strlen(s1);
-        x3::parse(s1, e1, '{' >> int_[std::bind(fun_action(), _1)] >> '}');
+        x3::parse(s1, e1, '{' >> int_[boost::bind(fun_action(), _1)] >> '}');
     }
 
     BOOST_TEST(x == (42*3));
--- a/test/x3/kleene.cpp
+++ b/test/x3/kleene.cpp
@@ -90,20 +90,16 @@ main()
     }
 
     { // actions
-        using boost::spirit::x3::_attr;
-
         std::string v;
-        auto f = [&](auto& ctx){ v = _attr(ctx); };
+        auto f = [&](auto& ctx){ v = boost::spirit::x3::_attr(ctx); };
 
         BOOST_TEST(test("bbbb", (*char_)[f]) && 4 == v.size() &&
             v[0] == 'b' && v[1] == 'b' && v[2] == 'b' &&  v[3] == 'b');
     }
 
     { // more actions
-        using boost::spirit::x3::_attr;
-
         std::vector<int> v;
-        auto f = [&](auto& ctx){ v = _attr(ctx); };
+        auto f = [&](auto& ctx){ v = boost::spirit::x3::_attr(ctx); };
 
         BOOST_TEST(test("123 456 789", (*int_)[f], space) && 3 == v.size() &&
             v[0] == 123 && v[1] == 456 && v[2] == 789);
--- a/test/x3/list.cpp
+++ b/test/x3/list.cpp
@@ -89,10 +89,8 @@ main()
     }
 
     { // actions
-        using boost::spirit::x3::_attr;
-
         std::string s;
-        auto f = [&](auto& ctx){ s = std::string(_attr(ctx).begin(), _attr(ctx).end()); };
+        auto f = [&](auto& ctx){ s = std::string(boost::spirit::x3::_attr(ctx).begin(), boost::spirit::x3::_attr(ctx).end()); };
 
         BOOST_TEST(test("a,b,c,d,e,f,g,h", (char_ % ',')[f]));
         BOOST_TEST(s == "abcdefgh");
--- a/test/x3/uint1.cpp
+++ b/test/x3/uint1.cpp
@@ -131,12 +131,11 @@ main()
     //  action tests
     ///////////////////////////////////////////////////////////////////////////
     {
-        using boost::spirit::x3::_attr;
         using boost::spirit::x3::uint_;
         using boost::spirit::x3::ascii::space;
         int n;
 
-        auto f = [&](auto& ctx){ n = _attr(ctx); };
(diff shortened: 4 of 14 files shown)


djowel:
Can MSVC handle all the tests and examples with this fix?

hia3:
No. I have only 30 tests of x3 passing. 18 tests are still failing to compile. But rexpr_min works.

Examples build successfully: x3_calc1.exe, x3_calc4.exe, x3_calc8.exe, x3_calc9.exe, x3_employee.exe, x3_num_list1.exe, x3_num_list4.exe, x3_rexpr.exe.

I think there is like 1 or 2 more workarounds to do.

For example:

spirit/home/x3/core/call.hpp(57): error C2672: "operator __surrogate_func": overloaded function not found

hia3:
This fixes all MSVC2015-specific problems with the library itself. All tests are passing, all examples are compiling (except one test and one example, see below)

There are still 2 issues:

1) rexpr\rexpr_full\test\parse_rexpr_test.cpp(38): error C2664: unable to convert argument 4 from "const boost::filesystem::path::value_type *" to "std::string"

This is probably boost::filesystem's defect (if it allows this same code to compile on another platforms).

2) x3_char1-p3.test:
boost/spirit/home/x3/char/any_char.hpp(57): error C2398: Element 1 and 2: narrowing conversion from "const wchar_t" to "char".

I've tried clang - it gives warning about this line. This warning/error should be fixed, but I don't know how to convert wchar_t to char.

test/x3/list.cpp
```diff
@@ -89,10 +89,8 @@ main()
     }
 
     { // actions
-        using boost::spirit::x3::_attr;
-
         std::string s;
-        auto f = [&](auto& ctx){ s = std::string(_attr(ctx).begin(), _attr(ctx).end()); };
+        auto f = [&](auto& ctx){ s = std::string(boost::spirit::x3::_attr(ctx).begin(), boost::spirit::x3::_attr(ctx).end()); };
```
djowel: This is not idiomatic. Does it not work without the full qualification?

test/x3/kleene.cpp
```diff
@@ -90,20 +90,16 @@ main()
...
         BOOST_TEST(test("bbbb", (*char_)[f]) && 4 == v.size() &&
             v[0] == 'b' && v[1] == 'b' && v[2] == 'b' &&  v[3] == 'b');
     }
 
     { // more actions
-        using boost::spirit::x3::_attr;
-
         std::vector<int> v;
-        auto f = [&](auto& ctx){ v = _attr(ctx); };
+        auto f = [&](auto& ctx){ v = boost::spirit::x3::_attr(ctx); };
```
djowel: Why do you need full qualification?

test/x3/uint1.cpp
```diff
@@ -131,12 +131,11 @@ main()
     //  action tests
     ///////////////////////////////////////////////////////////////////////////
     {
-        using boost::spirit::x3::_attr;
         using boost::spirit::x3::uint_;
         using boost::spirit::x3::ascii::space;
         int n;
 
-        auto f = [&](auto& ctx){ n = _attr(ctx); };
+        auto f = [&](auto& ctx){ n = boost::spirit::x3::_attr(ctx); };
```
djowel: Same here

test/x3/actions.cpp
```diff
@@ -65,9 +67,9 @@ int main()
     }
 
     {
-        using namespace std::placeholders;
+        using namespace boost::placeholders;
         char const *s1 = "{42}", *e1 = s1 + std::strlen(s1);
-        x3::parse(s1, e1, '{' >> int_[std::bind(fun_action(), _1)] >> '}');
+        x3::parse(s1, e1, '{' >> int_[boost::bind(fun_action(), _1)] >> '}');
```
djowel: Why change std::bind to boost::bind?

test/x3/actions.cpp
```diff
@@ -65,9 +67,9 @@ int main()
     }
 
     {
-        using namespace std::placeholders;
+        using namespace boost::placeholders;
```
djowel: Why change to boost?

test/x3/list.cpp
hia3: Because (without the full qualification) it results in internal compiler error. I just wanted to fix as much tests as I could. - I'm ok with rolling this change back as it is just a test.

test/x3/actions.cpp
hia3: With std::bind it does not compile, it static_assert-s "tuple index out of bounds". I've tried to remove this static_assert - compiles, works. Can rollback.

hia3:
And btw, this - https://github.com/boostorg/spirit/blob/547076384ac4d9d6c94298015119c759f02c4882/doc/x3/tutorial/actions.qbk#L63 should say we can omit the parameter.

djowel:
The basic problem with MSVC is the fragility of the compiler and support for c++14. Things like what you did can easily trigger internal compiler errors. To be honest, I'm reluctant to support the compiler, considering that further development can easily disrupt the code base due to this fragility. I am not sure I am convinced yet that MSVC is good enough.

hia3:
Who is sayng "support"? Just let it work if it works. Should I remove changes in x3 tests?

djowel:
either way, i'm not sure. let me think about it. x3 is supposed to be workaround free.

hia3:
Ok, I see. )

djowel:
give me some time, please. i DO value your patch.

unjello:
:+1: Hi there. Any chance these fixes will get to mainline? I hate staying on patched versions :)

djowel:
I'm very sorry, but I'll have to say no on this one :-( 

When MSVC supports C++14, then X3 will support MSVC meanwhile ... please use V2 (Qi). There will be more c++14-isms coming that will break more of the MSVC code. I don't want MSVC to tie us down to what idioms we will use. Thank you for your concern about Spirit running on MSVC. I think your energy will be best directed at the Microsoft development team in helping them produce a compliant compiler.

unjello:
Well... I would like to be able to persuade you, but the truth is I totally get your point. Just checked the second preview of latest MSVC, v15.0.25302.0 PREVIEW - and they still didn't fix this :/

rubenvb:
Perhaps it's more valuable to support MSVC's Clang frontend instead? That would mean a decent compiler and a different standard library.

CaseyCarter:
> Perhaps it's more valuable to support MSVC's Clang frontend instead? That would mean a decent compiler and a different standard library.

Clang/C2 uses Microsoft's STL.

mxc-commons:
Some german Dax enterprises we are working for decided strategically for Microsoft. So I'd love to see MSVC support.

timangus:
Would you consider committing the BOOST_SPIRIT_DEFINE fix only (c51a013)? This at least makes it compile and be usable on VS2015, even if some of the tests fail. I was just about to submit a basically identical patch that I came up with independently. I submitted a bug report for the compiler too, FWIW: https://connect.microsoft.com/VisualStudio/feedback/details/3109357/cannot-overload-templated-functions-with-decltype-parameters

unjello:
I'm more than happy to report, that `VS/15.0.0-RC.4+26206.0`'s platform toolset handles (at least) this case :) 👍

djowel:
Awesome!

Kojoley:
I am closing the PR because it had grown with merge conflicts and was relevant to a beta compiler (according to @unjello the problem solved in RC).

Kojoley:
Replaced by #308

PR #169, opened 2016-01-25
Added support for std::multimap, std::unordered_map and std::unordered_multimap.

octopus-prime:
Added container traits.
Added test cases.


include/boost/spirit/home/x3/support/traits/container_traits.hpp
```diff
@@ -118,6 +119,24 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
...
+        {
+            c.insert(std::move(val));
+        }
+
+        template <typename Key, typename Value, typename Hash, typename KeyEqual, typename Allocator, typename T>
+        static void push_back(std::unordered_multimap<Key, Value, Hash, KeyEqual, Allocator>& c, T&& val)
+        {
+            c.insert(std::move(val));
+        }
+
```
djowel: There should be a way to generalize this for all containers that support c.insert(x), I think. That way, we don't have to include the files and still allow for other containers.

octopus-prime:
Good idea. Work is done. Tests are extended.

djowel:
wonderful.

PR #170, opened 2016-01-26
Cleaned up container_traits.

octopus-prime:
Cleaned up push_back_container and append_container.

Removed reserve optimization in append_container since we must provide reserve for different container types (e.g. string and deque) and the container implementation should decide how to append elements as fast as possible.

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/container_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/container_traits.hpp
@@ -149,9 +149,26 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
     template <typename Container, typename Enable = void>
     struct append_container
     {
+        // Not all containers have "reserve"
+        template <typename Container_>
+        static void reserve(Container_& c, std::size_t size) {}
+
+        template <typename T, typename Allocator>
+        static void reserve(std::vector<T, Allocator>& c, std::size_t size)
+        {
+            c.reserve(size);
+        }
+
+        template <typename T, typename Traits, typename Allocator>
+        static void reserve(std::basic_string<T, Traits, Allocator>& c, std::size_t size)
+        {
+            c.reserve(size);
+        }
+
         template <typename Iterator>
         static bool call(Container& c, Iterator first, Iterator last)
         {
+        	reserve(c, c.size() + std::distance(first, last));
         	c.insert(c.end(), first, last);
             return true;
         }
--- a/include/boost/spirit/home/x3/support/traits/container_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/container_traits.hpp
@@ -112,15 +112,9 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
     struct push_back_container
     {
         template <typename T>
-        static void push_back(Container& c, T&& val)
-        {
-            c.insert(c.end(), std::move(val));
-        }
-       
-        template <typename T>
         static bool call(Container& c, T&& val)
         {
-            push_back(c, std::move(val));
+            c.insert(c.end(), std::move(val));
             return true;
         }
     };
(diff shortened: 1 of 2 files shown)


djowel:
Oops. We need that optimization... please to find another way.

octopus-prime:
Hm... How to provide reserve for so many container types?

string, vector, deque, unordered_map, unordered_multimap, unordered_set, unordered_multiset...

octopus-prime:
Reserve optimization for most important types.

djowel:
yes, we'll have to do some special cases here. that's necessary for optimization.

octopus-prime:
Do you have some simple grammars for testing container traits using append?

hkaiser:
@octopus-prime What's your rationale for this change? Shouldn't we rather provide specializations for those container types you're mentioning?

octopus-prime:
@hkaiser Look at pull request #169. There we removed specializations from push_back_container and used the insert() method witch all common containers provide. So we have less code and we provide more container types - out of the box.

The basic idea in pull request #170 was the same for append_container. To remove specializations and use the insert() method witch all common containers provide. You know providing specializations for each container type on earth is endless work with a lot of spaghetti specialization code...

But djowel told me that we need the reserve() optimization. So the the basic idea is gone...
We need specializations. Perhaps we can close this pull request.

Nevertheless, we should have tests for the append_container. So i need simple grammars for testing container traits using append.

djowel:
The reserve optimization *IS* important. The tricky part is how to generalize this for all containers. For example, we can use a metaprogram that detects the availability of the reserve member function for a container C.

octopus-prime:
Done. Please do an intense code review!!

djowel:
Where's the "like" button? Boost folks are smart! :-)

include/boost/spirit/home/x3/support/traits/container_traits.hpp
```diff
@@ -155,27 +165,21 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
...
+            c.reserve(size);
         }
 
         template <typename Iterator>
         static bool call(Container& c, Iterator first, Iterator last)
         {
-            reserve(c, c.size() + std::distance(first, last));
-            insert(c, first, last);
+        	reserve(c, c.size() + std::distance(first, last), std::integral_constant<bool, detail::has_reserve_method<Container>::value>());
+        	c.insert(c.end(), first, last);
```
tomilov: c.end() returns iterator, c.insert accept const_iterator. There is unneeded conversion.

include/boost/spirit/home/x3/support/traits/container_traits.hpp
```diff
@@ -155,27 +165,21 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
...
-            std::copy(first, last, std::inserter(c, c.end()));
+            c.reserve(size);
         }
 
         template <typename Iterator>
         static bool call(Container& c, Iterator first, Iterator last)
         {
-            reserve(c, c.size() + std::distance(first, last));
-            insert(c, first, last);
+        	reserve(c, c.size() + std::distance(first, last), std::integral_constant<bool, detail::has_reserve_method<Container>::value>());
```
tomilov: unsigned + signed. Is there warning issue?

tomilov: It better to use `typename detail::has_reserve_method<Container>::type{}` then `std::integral_constant<bool, detail::has_reserve_method<Container>::value>()`.

octopus-prime:
@tomilov Thanks!

- c.cend() ... done.
- static_cast< std::size_t >(std::distance(first, last)) ... i am not sure.
- typename detail::has_reserve_method<Container>::type{} ... looks very nice! done.

tomilov:
Defenitely there is an issue with multipass guarantee absence for input iterators. For them you cannot call distance, then read the same range once again. One can get logical (or even hard) error when using std::vector+std::istream_iterator combination.

octopus-prime:
From Qi documentation:

Backtracking in Spirit.Qi requires the use of the following types of iterator: forward, bidirectional, or random access. Because of backtracking, input iterators cannot be used. Therefore, the standard library classes std::istreambuf_iterator and std::istream_iterator, that fall under the category of input iterators, cannot be used.

Lack of X3 documentation here, but I think the same for X3.

octopus-prime:
If you try to use std::istream_iterator<char> as iterator type in x3::parse() you get:

error: conversion from ‘std::input_iterator_tag’ to non-scalar type ‘std::forward_iterator_tag’ requested

octopus-prime:
Some progress in merging this pull request?

djowel:
allow me to do another review this weekend.

octopus-prime:
In container_traits.hpp we have

template <typename Container, typename T>
inline bool push_back(Container& c, T&& val)

and

template <typename T>
inline bool push_back(unused_type, T const&)

I think there should be symmetry. So we should change the latter to

template <typename T>
inline bool push_back(unused_type, T&&)

?

Currently we can't do

x3::traits::push_back(attr, std::make_pair(key, value));

if attr is an unused_type.

djowel:
Unless there is an objection (Hartmut?), I think this is good to go.

octopus-prime:
Using end() because I got compiler errors with gcc 5.3 on Ubuntu 14.04.
The include files there want still iterator and not const_iterator.
end() should discard any of these problems.

hkaiser:
LGTM!

octopus-prime:
gcc 5.2.1 on Ubuntu 15.10 can build.
gcc 5.3.0 on Ubuntu 14.04 can build.

PR #174, opened 2016-02-23
Possible minimal fix for Trac #12016

sehe:
https://svn.boost.org/trac/boost/ticket/12016

It strikes me that moving from mapped values in a symbols trie is never a good idea, so the attribute value could just be marked const.

> _I'm not sure in this suggestion whether `handles_container` needs to be using `T` (or apply `remove_const<>` again...). Perhaps the trait already correctly identifies const types as containers._

### DIFF
--- a/include/boost/spirit/home/x3/string/symbols.hpp
+++ b/include/boost/spirit/home/x3/string/symbols.hpp
@@ -44,11 +44,11 @@ namespace boost { namespace spirit { namespace x3
     {
         typedef typename Encoding::char_type char_type; // the character type
         typedef Encoding encoding;
-        typedef T value_type; // the value associated with each entry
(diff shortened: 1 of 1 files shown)


djowel:
Beat you to a solution, sehe :-) See fix in develop (https://github.com/boostorg/spirit/commit/035fc0fd9d511d4819171dd9af3d740ba767ed00)

Tell me if that works for you. 

PS> There are times when you want the data non-const (e.g. the nabialek trick).

sehe:
That's a lot more to the point yes. 

Why does the `move_to` "trait" move from lvalue references, though? Seems error-prone. I bet there is a good reason, but I assumed it would be preferable to perfect-forward rvalue-ness of all arguments?

djowel:
Yes, I've been thinking about that as well (move from lvalue references). I need to investigate more on this.

PR #176, opened 2016-03-16
[MSVC] problem with BOOST_FUSION_DEFINE_STRUCT_INLINE

hia3:
There is a bug in MSVC that it will call non-explicit constructor of a destination type instead of calling type conversion operator of source class.

BOOST_FUSION_DEFINE_STRUCT_INLINE generates a class that has non-explicit template constructor calling boost::fusion::copy. MSVC instantianes it and fails in boost::fusion::copy because boost::value_initialized<T> is not a sequence.

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/value_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/value_traits.hpp
@@ -18,7 +18,7 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
     {
         static T call()
         {
-            return boost::value_initialized<T>();
+            return boost::value_initialized<T>().data();
         }
     };
 }}}}


daminetreg:
Making explicit default constructors in BOOST_FUSION_DEFINE_STRUCT* wouldn't solve the issue better ?

hia3:
Making this https://github.com/boostorg/fusion/blob/develop/include/boost/fusion/adapted/struct/detail/define_struct_inline.hpp#L442 constructor explicit will solve the problem. Even though making constructors explicit is a good thing in general, I have no idea if is better in this case.

daminetreg:
Oh this one has to stay implicit, as it allows implicit conversion. One that however be explicit is https://github.com/boostorg/fusion/blob/develop/include/boost/fusion/adapted/struct/detail/define_struct_inline.hpp#L427, but this won't solve your problem in this case.

So sorry for disturbing there. :smile:

hia3:
Okay, how about enable_if-ing it?

```
    template <typename boost_fusion_detail_Seq, typename e = boost::enable_if<typename boost::fusion::traits::is_sequence<boost_fusion_detail_Seq>::type, void>::type>                                 \
    BOOST_CXX14_CONSTEXPR BOOST_FUSION_GPU_ENABLED                              \
    NAME(const boost_fusion_detail_Seq& rhs)                                    \
    {                                                                           \
        boost::fusion::copy(rhs, *this);                                        \
    }                                                                           \

```

Kojoley:
All x3 tests pass on MSVC 2015/2017.
Is this still an issue?

PR #177, opened 2016-03-19
fixes problem with move_to

hia3:
Fix bug described here - http://article.gmane.org/gmane.comp.lib.boost.devel/266211

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/move_to.hpp
+++ b/include/boost/spirit/home/x3/support/traits/move_to.hpp
@@ -23,7 +23,19 @@
 namespace boost { namespace spirit { namespace x3 { namespace traits
 {
     template <typename Source, typename Dest>
-    void move_to(Source&& src, Dest& dest);
+    inline void move_to(Source&& src, Dest& dest);
+
+    template <typename T>
+    inline void move_to(T& src, T& dest);
+
+    template <typename T>
+    inline void move_to(T const& src, T& dest);
+
+    template <typename T>
+    inline void move_to(T&& src, T& dest);
+
+    template <typename Iterator, typename Dest>
+    inline void move_to(Iterator first, Iterator last, Dest& dest);
 
     template <typename Dest>
     inline void move_to(unused_type, Dest&) {}
@@ -32,10 +44,6 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
     inline void move_to(Source&, unused_type) {}
 
     inline void move_to(unused_type, unused_type) {}
-
-    template <typename Iterator, typename Dest>
-    void
-    move_to(Iterator first, Iterator last, Dest& dest);
 
     template <typename Iterator>
     inline void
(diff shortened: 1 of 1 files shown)


djowel:
Please explain the "problem with move_to" and how this fixes it. I'll also need some tests the exhibit the problem you mention.

hia3:
Should I explain it here in comments of in commit description? Should I add example from http://article.gmane.org/gmane.comp.lib.boost.devel/266211 to the commit?

hia3:
Problem was that it was trying to call this traits::move_to https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/x3/support/traits/move_to.hpp#L171 for (in my case) Source = a1 and Dest = a1.

Message from compiler:
/..../spirit-develop/include/boost/spirit/home/x3/support/traits/move_to.hpp:172:17: note: in instantiation of function template specialization 'boost::spirit::x3::traits::detail::move_to<ast::a1, ast::a1>' requested here

Patch just adds all the forward declarations of traits::move_to so that they are vidible from detail namespace.

hia3:
I'm not sure all those forwarding references and `std::move`s are necessary. Because 1) it can be `std::move`d when real work is done (in assignment for example) and 2) `default_transform_attribute::post` gets `attr` by reference anyway.

djowel:
move_to in develop is work in progress. i'd be more comfortable if you can just provide a minimal test case that exhibits the problem you mention.

hia3:
Isn't it minimal enough?

    #include <boost/spirit/home/x3/support/ast/variant.hpp>
    #include <boost/fusion/include/define_struct_inline.hpp>

    #include <string>
    #include <vector>

    namespace ast
    {
        BOOST_FUSION_DEFINE_STRUCT_INLINE
        (
            a1,
            (std::string, v)
        )

        BOOST_FUSION_DEFINE_STRUCT_INLINE
        (
            a2,
            (a1, v1)
            (std::vector<double>, v2)
        )

        BOOST_FUSION_DEFINE_STRUCT_INLINE
        (
            a3,
            (std::vector<a2>, v1)
        )
    }

    auto a0 = rule<class i0, std::string>{} = lexeme[+alpha];
    auto a1 = rule<class i1, ast::a1>{} = a0;

    auto a2_v2 = rule<class a2_v2, std::vector<double>>{} = -(double_ % ',');
    auto a2 = rule<class i2, ast::a2>{} = a1 >> '(' >> a2_v2 >> ')';

    auto a3_v1 = rule<class a3_v1, std::vector<ast::a2>>{} = *(a2 >> ';');
    auto a3 = rule<class i3, ast::a3>{} = a3_v1;

    using data = ast::a3;

    int main()
    {
        std::string script = "";

        auto begin = script.begin();
        auto   end = script.end();

        data v;

        bool ok = (boost::spirit::x3::phrase_parse(begin, end, a3, space, v) && begin == end);
    }

djowel:
perfect!

djowel:
Ok, I'm going to merge this patch. Seems it's not in conflict with the work I am doing.

PR #178, opened 2016-03-19
X3: Kleene star can't into single-element sequence

hia3:
This will fail to compile with error "'value_type': is not a member of 'ast::a3'".

```cpp
#include <boost/spirit/home/x3/support/ast/variant.hpp>
#include <boost/fusion/include/define_struct_inline.hpp>

#include <string>
#include <vector>

namespace ast
{
    BOOST_FUSION_DEFINE_STRUCT_INLINE
    (
        a1,
        (std::string, v)
    )

    BOOST_FUSION_DEFINE_STRUCT_INLINE
    (
        a2,
        (a1, v1)
        (std::vector<double>, v2)
    )

    BOOST_FUSION_DEFINE_STRUCT_INLINE
    (
        a3,
        (std::vector<a2>, v1)
    )
}

auto a1 = rule<class i1, ast::a1>{} = lexeme[+alpha];

auto a2 = rule<class i2, ast::a2>{} = a1 >> '(' >> -(double_ % ',') >> ')';

auto a3 = rule<class i3, ast::a3>{} = *(a2 >> ';');

using data = ast::a3;

int main()
{
    std::string script = "foo(1,2,3);";

    auto begin = script.begin();
    auto   end = script.end();

    data v;

    bool ok = (boost::spirit::x3::phrase_parse(begin, end, a3, space, v) && begin == end);
}
```

Fixed by adding container_value specialization: for single-element fusion sequence return container_value of that one element.

---
EDIT by maintainer: fixed styling issues

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/container_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/container_traits.hpp
@@ -11,6 +11,7 @@
 
 #include <boost/fusion/support/category_of.hpp>
 #include <boost/spirit/home/x3/support/unused.hpp>
+#include <boost/spirit/home/x3/support/traits/tuple_traits.hpp>
 #include <boost/detail/iterator.hpp>
 #include <boost/fusion/include/deque.hpp>
 #include <boost/mpl/has_xxx.hpp>
@@ -62,20 +63,20 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
             typedef std::pair<first_type, second_type> type;
         };
 
-    	template <class Container>
-    	struct has_reserve_method
-    	{
-    	    template <class T>
-    	    static std::true_type test_signature(void (T::*)(std::size_t));
+        template <class Container>
+        struct has_reserve_method
+        {
+            template <class T>
+            static std::true_type test_signature(void (T::*)(std::size_t));
 
-    	    template <class T>
-    	    static decltype(test_signature(&T::reserve)) test(std::nullptr_t);
+            template <class T>
+            static decltype(test_signature(&T::reserve)) test(std::nullptr_t);
 
-    	    template <class T>
-    	    static std::false_type test(...);
+            template <class T>
+            static std::false_type test(...);
 
-    	    using type = decltype(test<Container>(nullptr));
-    	};
+            using type = decltype(test<Container>(nullptr));
+        };
     }
 
     ///////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


hia3:
Btw, adding #define BOOST_SPIRIT_X3_DEBUG will break this sample program again.

djowel:
I'm not sure about this one. Why do you have to wrap the container into a fusion vector? Why not simply use std::vector<double> and std::vector<std::vector<double>> ?

hia3:
Why not?

I'm not saying "you have to". This might be the result of simplification of some existing code. Do you want to not to compile simpler version of the code? Then `static_assert` would help a lot. Or this might be the 1st parser someone new to the library (like myself) trying to write.

If I'd adapting the existring struct, I could write `using value_type = whatever;`, but with `BOOST_FUSION_DEFINE_STRUCT_INLINE` there is no way I can do it.

Why `variable` here https://github.com/boostorg/spirit/blob/develop/example/x3/calc/calc9/ast.hpp#L27 is not a string?

hia3:
Btw, `ast::a3` is unchanged from my previous pull request (it was single element seqence too). Only thing changed is a3 rule from this:

    auto a3_v1 = rule<class a3_v1, std::vector<ast::a2>>{} = *(a2 >> ';');
    auto a3    = rule<class i3, ast::a3>{}                 = a3_v1;

to this

    auto a3    = rule<class i3, ast::a3>{}                 = *(a2 >> ';');

And now it suddenly requires `value_type` to be defined.

djowel:
I'll have to think about this some more as it introduces a new behavior. The previous behavior in which you provided a patch for is suspicious to begin with. What we should probably do is to distill the test code to the bare essentials so I (we) can analyze and reason out with the correct behavior. I fear that we are opening a can of worms if we allow single-element-tuple-with-container == container.

Kojoley:
MWE:

```cpp
#include <boost/fusion/include/define_struct_inline.hpp>
#include <string>

namespace ast
{
    BOOST_FUSION_DEFINE_STRUCT_INLINE
    (
        a3,
        (std::vector<int>, v1)
    )
}

int main()
{

    auto test = [](auto&& p)
    {
        std::string script = "foo(1,2,3);";

        auto begin = script.begin();
        auto   end = script.end();

        ast::a3 v;

        return (x3::phrase_parse(begin, end, p, x3::space, v) && begin == end);
    };

    test(x3::rule<class tst_1_r, ast::a3>{} = *x3::int_);  // OK
    test(x3::rule<class tst_2_r, ast::a3>{} = x3::int_ % ';'); // OK
    test(x3::rule<class tst_3_r, ast::a3>{} = *(x3::int_ >> ';')); // Fail to compile
}
```

saki7:
The code shown above seems to work in 2025. We already have tests that consist of same combination of grammars: e.g. https://github.com/boostorg/spirit/blob/328d940c372367a8d76a48ed956e870b56738ce9/test/x3/expect.ipp#L749

PR #179, opened 2016-03-26
dangling reference to attribute when BOOST_SPIRIT_X3_DEBUG is defined

hia3:
dbg_attribute_type(attr_) creates a temporary that is bound to const reference inside of context_debug.

This will fail on assert:

    #define BOOST_SPIRIT_X3_DEBUG

    #include <iostream>

    struct p
    {
        bool alive = true;

        void access() const
        {
            assert(alive);
        }
        ~p()
        {
            alive = false;
        }
    };

    std::ostream & operator << (std::ostream & os, p o)
    {
        o.access();
        return os << "o";
    }

    x3::symbols<p> s
    {
        {
            { "s", p{} }
        },
        "s"
    };

    auto r = x3::rule<struct r, p>{ "r" } = s;

    int main()
    {
        std::string str = "s";
        auto b = str.begin();
        auto e = str.end();

        p v;

        phrase_parse(b, e, r, x3::blank, v) && b == e;
    }

### DIFF
--- a/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp
@@ -325,7 +325,7 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
              // traits::post_transform when, for example,
              // ActualAttribute is a recursive variant).
 #if defined(BOOST_SPIRIT_X3_DEBUG)
-                typedef typename make_attribute::value_type dbg_attribute_type;
+                typedef typename make_attribute::type dbg_attribute_type;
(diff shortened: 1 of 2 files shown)


djowel:
This is a critical bug. Could you please add this test case in the x3 test suite as well?

hia3:
Added test case gives me

```
<b>
  <try>a</try>
  <success></success>
  <attributes>debug.cpp(43): test 'alive' failed in function 'void __thiscall my_attribute::access(void) const'
my_attribute</attributes>
</b>
1 error detected.

```

djowel:
Thanks!

PR #180, opened 2016-03-27
Test bug 12085

hia3:
new test for not_predicate.cpp

### DIFF
--- a/test/x3/char_class.cpp
+++ b/test/x3/char_class.cpp
@@ -239,9 +239,8 @@ main()
 
     {   // test action
         using namespace boost::spirit::x3::ascii;
-        using boost::spirit::x3::_attr;
         char ch;
-        auto f = [&](auto& ctx){ ch = _attr(ctx); };
+        auto f = [&](auto& ctx){ ch = boost::spirit::x3::_attr(ctx); };
 
         BOOST_TEST(test("x", alnum[f]));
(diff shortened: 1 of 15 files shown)


test/x3/char_class.cpp
```diff
@@ -239,9 +239,8 @@ main()
 
     {   // test action
         using namespace boost::spirit::x3::ascii;
-        using boost::spirit::x3::_attr;
         char ch;
-        auto f = [&](auto& ctx){ ch = _attr(ctx); };
+        auto f = [&](auto& ctx){ ch = boost::spirit::x3::_attr(ctx); };
```
djowel: Why do you have to do this?

test/x3/difference.cpp
```diff
@@ -43,20 +43,18 @@ main()
...
     {
-        using boost::spirit::x3::_attr;
-
         std::string s;
 
         BOOST_TEST(test(
             "/*abcdefghijk*/"
-          , "/*" >> *(char_ - "*/")[([&](auto& ctx){ s += _attr(ctx); })] >> "*/"
+          , "/*" >> *(char_ - "*/")[([&](auto& ctx){ s += boost::spirit::x3::_attr(ctx); })] >> "*/"
         ));
```
djowel: Please avoid doing this. I prefer the using.

PR #187, opened 2016-05-06
Fix compile error

phbaer:
This compile issue was introduced in commit e34a955. See also  comment of MarcelRaad (2 Mar 2015) for that commit.

### DIFF
--- a/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
+++ b/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
@@ -357,7 +357,7 @@ namespace boost { namespace spirit { namespace repository { namespace karma
         locals_type;
 
         typedef typename
-            spirit::detail::extract_sig<template_params>::type
+            spirit::detail::extract_sig<template_params, locals_type, karma::domain>::type
         sig_type;
 
         // This is the subrule's attribute type


djowel:
Aren't those defaulted?

phbaer:
Don't think so (https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/support/nonterminal/extract_param.hpp). Just got the compile error on MSVC2015 with Boost 1.60.0.

mlvdm:
Just want to say, this also seems to be the case in Qi as shipped in latest stable (and same fix seems to work, except replace karma with qi in the file path and the changed line). Also MSVC2015, if that matters. For sure extract_sig doesn't have default template params, so something needs to be passed.

djowel:
What strikes me as odd is that we're not getting these errors on other compilers. Our tests should've complained loudly.

mlvdm:
I suspect it's a case of MSVC parsing a bit too aggressively in a non-standard way. Since MSVC is not actually a two-phase compiler, I suspect that during parse it already sees no way that this could work and thus signals compilation issue. A "proper" compiler would only detect this in the second phase (ie, actual instantiation of subrule), which never happens in the code.

For example, if I add just below the `extract_sig` a `static_assert(false, "oops")`, this will fail at the same point (since false is non-dependent). However, if I add a `static_assert(ID_ != ID_, "oops")` it will compile fine (since ID_ is dependent), proving that the template is never actually instantiated. The other hint in the same direction is that the compilation error on `extract_sig` actually doesn't show a "stack" of instantiations to subrule that you would typically see in a template-heavy compiler error message.

For what it's worth, I could also reproduce this with clang/C2, but this may be because it's doing a bunch of MSVCisms as well.

I think the "other" way to work around this is to not include spirit/repository/home/qi.hpp or nonterminal.hpp, (thus avoiding including subrule.hpp) and the problem is also gone. This is what I'm doing currently, as I don't actually need subrule at all.

MarcelRaad:
It's the same compile error in the karma_repo_subrule and qi_repo_subrule tests for all compilers, not only MSVC:
http://www.boost.org/development/tests/develop/developer/spirit-repository.html

phbaer:
@mlvdm, thanks for looking into that issue as well. Your second workaround sounds good, we also don't use subrules. However, the issue remains.

As @MarcelRaad pointed out, it seems that the tests are actually failing. Are these new tests? To be honest, I did not check that state before.

djowel:
OK, can we do a complete PR for both Qi and Karma then?

Kojoley:
I think that this PR solves one problem, but adds other bug, because the second parameter for `extract_sig` is an encoding type, but not a local type.

I have tried to come up with the right fix, but I am not sure what it should be. The simplest thing we could done about that is something like this 3879894ca8b2056230dfade89cc25b4deb45adbf (tested with GCC 5.0 and clang 3.8), but maybe encoding support has to be added to the `subrule` like it was done in c5c871c89240bc5937f4c0082da17bb8b73eda68 to the `rule`?

Kojoley:
Superseded by #259

PR #188, opened 2016-05-12
Suppress several 'unused parameter' warnings

bebuch:


include/boost/spirit/home/x3/nonterminal/detail/rule.hpp
```diff
@@ -328,6 +328,9 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
                 typedef typename make_attribute::type dbg_attribute_type;
                 context_debug<Iterator, dbg_attribute_type>
                 dbg(rule_name, first, last, dbg_attribute_type(attr_), ok_parse);
+#else
+                // supress unused parameter warning
+                (void)rule_name;
```
Kojoley: Maybe it's better to use [`boost::ignore_unused(rule_name)`](http://www.boost.org/doc/libs/1_60_0/libs/core/doc/html/core/ignore_unused.html) ?

bebuch:
I added the ignore_unused()-function, it is more comprehensible.

PR #190, opened 2016-05-16
Simple tests for istream_iterator and a fix for multi_pass_iterator

jefftrull:
This should resolve TRAC #12166 and add some test coverage for istream_iterator (including the fix).


bebuch:
The relational operators in lines 155, 159 and 163 can also be const.

jefftrull:
@bebuch very true :)  I just wasn't sure what their purpose was and decided to leave them alone...

PR #191, opened 2016-05-17
Repair plain_token_range and add tests

jefftrull:
plain_token_range has a bug where it really only checks for equality against the bottom of the range supplied.  This patch restores the intended behavior.


djowel:
Would it make sense to have a test for this? I might have to defer this one to @hkaiser. Thoughts?

hkaiser:
LGTM! thanks!

PR #192, opened 2016-05-17
add const to relational operators of class template multi_pass

bebuch:


djowel:
multi_pass is an odd beast. semi-random-access, semi-forward.

PR #193, opened 2016-05-31
Remove auto_ptr (deprecated in recent C++ standards).

Romain-Geissler-1A:
Hi,

I am having issues when using json_spirit and boost 1.61 with GCC 6 and -std=gnu++14 -Wall -Werror -Wextra.

This removes all the last auto_ptr references in boost spirit. Inspired by https://svn.boost.org/trac/boost/ticket/11672.

I have tested with json_spirit and boost 1.61, then ported to the branch "develop" as is.

Cheers,
Romain

### DIFF
--- a/include/boost/spirit/home/classic/core/non_terminal/impl/grammar.ipp
+++ b/include/boost/spirit/home/classic/core/non_terminal/impl/grammar.ipp
@@ -15,7 +15,7 @@
 #include <boost/spirit/home/classic/core/non_terminal/impl/object_with_id.ipp>
 #include <algorithm>
 #include <functional>
-#include <memory> // for std::auto_ptr
+#include <boost/move/unique_ptr.hpp>
 #include <boost/weak_ptr.hpp>
 #endif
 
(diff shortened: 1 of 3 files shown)


prlw1:
Question from the peanut gallery: why not go for std::unique_ptr from <memory> rather than boost::movelib::unique_ptr ?

Romain-Geissler-1A:
Most of boost library are meant to work on a wide range of platform/compilers/C++ standards. A non neglibeagle part of people in the industry still use C++03 (Amadeus which I am working for is currently massively migrating to C++11, and I am preparing the next move to C++14). AFAIK most the current Boost libraries currently supports that (using C++03). There are some exceptions like Boost Hana for example, but not a lot.

So I could have used a #ifdef to conditionally use std::unique_ptr when available, but for me adding a lot of #ifdef bloats the code. On the other hand, boost::movelib::unique_ptr is available no matter what configuration you use.

djowel:
That is not necessary. Spirit Classic is meant to work for older compilers as a backward compatible support for legacy Spirit code.

prlw1:
Thanks for the explanation! (I missed the obvious - not all the world is x3)

djowel:
My bad. I'll do so as soon as possible.

djowel:
merged to master

Romain-Geissler-1A:
Thank you. Is this too late to be included in Boost 1.65 ? Or will it be in 1.66 ?

PR #194, opened 2016-06-09
Spirit QI grammar doesn't work

tiagokepe:
I've followed the Spirit QI tutorial based on the bellow grammar (http://www.boost.org/doc/libs/1_60_0/libs/spirit/doc/html/spirit/introduction.html):

group       = '(' >> expression >> ')';
factor      = integer | group;
term        = factor >> *(('*' >> factor) | ('/' >> factor));
expression  = term >> *(('+' >> term) | ('-' >> term));

The parser returns true to the input: "1 blabla 2".

### DIFF
--- a/include/boost/spirit/home/classic/core/non_terminal/impl/grammar.ipp
+++ b/include/boost/spirit/home/classic/core/non_terminal/impl/grammar.ipp
@@ -15,7 +15,7 @@
 #include <boost/spirit/home/classic/core/non_terminal/impl/object_with_id.ipp>
 #include <algorithm>
 #include <functional>
-#include <memory> // for std::auto_ptr
+#include <boost/move/unique_ptr.hpp>
 #include <boost/weak_ptr.hpp>
 #endif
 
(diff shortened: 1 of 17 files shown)


vtnerd:
There appears to be a lot more commits in this pull request than necessary, and more than likely Joel will want this pull request _into_ develop.

Kojoley:
There are no commits at all. Just a question and I do not know why it is done as a pull request.

> group = '(' >> expression >> ')';
> factor = integer | group;
> term = factor >> (('' >> factor) | ('/' >> factor));
> expression = term >> *(('+' >> term) | ('-' >> term));

This grammar is ok and it must succeed with the "1 blabla 2" input. It will just parse "1" (expression -> term -> factor -> integer), you can check it by comparing you first iterator to the last (iter != last)

tiagokepe:
I disagree that "1 blabla 2" must succeed, because it is an expression that doesn't belong to the grammar. I thing that such expression must to be evaluated as a whole.

djowel:
It is a partial match. If you do not agree, please ask your question to the spirit list for discussion instead: http://boost-spirit.com/home/feedback-and-support/. I'm closing this one. If there are patches, please be specific and file them separately.

PR #195, opened 2016-06-29
Don't try to declare std::complex for Android builds

KayEss:
This breaks for clang as the NDK namespaces are oddly set up.

The old version works for gcc builds on Android, but clang is now the default compiler with gcc deprecated.

Isn't the declaration undefined behaviour anyway?


djowel:
Doh. That is indeed undefined behavior! Please update the PR fix by removing it entirely.

djowel:
Thanks!

PR #196, opened 2016-07-12
Expect directive for spirit qi 2

mxc-commons:
As discussed with Joel de Guzman this PR introduces the expect directive to
spirit qi 2. PR includes code, test, example and documentation.
See https://sourceforge.net/p/spirit/mailman/spirit-general/thread/nln5d3%24klg%241%40ger.gmane.org/#msg35208052

1. Added expect.hpp to home/qi/directives.
1a. Added forwarding header qi_expectd.hpp in include.
2. Renamed expectation parser (a > b) struct from expect to expect_operator
   to prevent name conflicts (expect directive uses struct expect_directive).
3. Added expect directive to the common terminals (common_terminals.hpp).
4. Added expectd.cpp to the test suite.
5. Added expectd.cpp to the examples.
6. Added expect directive documentation to directive.qbk. Crosslinked 
   expectation operator and expect directive docs.
7. Added expect directive to the index.idx file for index-generation.
8. Added expect directive to the qi quick_reference.qbk.
9. Added expect directive to the spirit reference card (tex, pdf)
10. Created a new section in the whats_new.hpp.
11. Changed spirit version to 2.5.4. Spirit index.html showed 2.5.2
    allthough an 2.5.3 What's New section already existed. Fixed.


include/boost/spirit/include/qi_expectd.hpp
```diff
@@ -0,0 +1,19 @@
...
+    Copyright (c) 2001-2011 Joel de Guzman
+    Copyright (c) 2001-2011 Hartmut Kaiser
+    Copyright (c) 2016      Frank Hein, maxence business consulting gmbh
+    http://spirit.sourceforge.net/
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================*/
+#ifndef BOOST_SPIRIT_INCLUDE_QI_EXPECTD
+#define BOOST_SPIRIT_INCLUDE_QI_EXPECTD
```
djowel: Why is this named expectd? It does not follow our naming conventions.

djowel:
BTW, Spirit X3 has the expect directive. It would be best to sync the naming and directory structure and implementation with that. Have you taken a look? So far everything is good, but I want to be sure we have consistency.

mxc-commons:
I'll check and change.

include/boost/spirit/include/qi_expectd.hpp
mxc-commons: It's called qi_expectd.hpp because there was a qi_expect.hpp in the include directory (for the expect operator). Please advice.

djowel: How about doing what XR does?

mxc-commons: I still do not understand what you mean. What does X3 differently? What am I missing?

mxc-commons: X3 seems not to have an expect operator.

vtnerd: The X3 expect operator (`operator>`) is in operator/sequence.hpp

mxc-commons: Oh, I see. Would be all but trivial to refactor the qi expect operator to act like x3's.

mxc-commons: So how to continue?

djowel: Your call. You can probably just add the new commits in this PR, no?

mxc-commons: I could and I would. I actually don't have a clue what you want me to do neither what the problem in particular is.

Is it that a header qi_expectd.hpp in include is not acceptable? We could just remove it. As far as I understand it, the current header structure is in home, and include headers are there for historical reasons forwarding to home. If that was true, i would remove the header and adjust the doc. Is it that what I should do?

Or should I forward include both directive and operator header from qi_expect.hpp? I could do that. This would make qi_expectd obsolete without dropping the forward include. 
I assume this to be the best solution because, so I will start by committing that.

Should I change the expect operators implementation? And how? Should it use the expect directive for the parsing job of the operator? IMO there is no benefit from that which would justify the change. And would introduce a new dependency.

Or is it at last the only the name of the guardian which is not ok?

Can you please point me to the naming conventions you referred to? "The Porting from 1.8" in the docs is the only I found mentioning naming conventions specific to qi. And, of course, the boost conventions which do not go into the detail of the library's internals.

djowel: Your "best solution" sounds good. There's actually no formal naming convention. Just follow what we have in the current code, e.g. no abbreviations such as expectd.

mxc-commons: Comitted already. :)

include/boost/spirit/home/qi/directive/expect.hpp
```diff
@@ -0,0 +1,115 @@
...
+    using spirit::expect;
+#endif
+    using spirit::expect_type;
+
+    template <typename Subject>
+    struct expect_directive : unary_parser<expect_directive<Subject> >
+    {
+        using subject_type 
+            = result_of::compile<domain, Subject>;
+
```
djowel: Ooops. Will this work with C++03? Have we checked if the additions work with C++03? That's our minimum for Qi and Karma (not so for X3).

djowel:
Oh, please test against C++03. I might have spotted some C++11-isms.

include/boost/spirit/home/qi/directive/expect.hpp
mxc-commons: I removed the type aliases. Sorry for that.

djowel:
Great! Are we good to go? Everything tested on C++03? Please confirm so I can merge.

mxc-commons:
Tested ok against MSVC++ 2006, 2008, 2013 and 2015.

PR #198, opened 2016-07-19
Fixes for PR #196 expect directive

mxc-commons:
his PR contains fixes for both issues reported by Lastique and
MarcelRaad. It makes PR #197 obsolete.

Removed obsolete dependency  of test/qi/expectd.cpp on
mxc/qitoo/qitoo.hpp. Was reintroduced by mistake.

Created qi/detail/expectation_failure.hpp. It defines the
expectation_failure template as a common dependency of expect directive
and expect operator. Removed expectation_failure from
operator/expect.hpp.

Changed dependency of debug_handler.hpp to from operator/expect.hpp to expectation_failure.hpp.

Changed version.hpp to 2.54. I did not notice that addl file in the
first place.

### DIFF
--- a/include/boost/spirit/home/qi/detail/expectation_failure.hpp
+++ b/include/boost/spirit/home/qi/detail/expectation_failure.hpp
@@ -11,6 +11,8 @@ file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 #pragma once
 #endif
 
+#include <stdexcept>
+
 namespace boost { namespace spirit { namespace qi {
     template <typename Iterator>
     struct expectation_failure : std::runtime_error
--- a/include/boost/spirit/home/qi/detail/expectation_failure.hpp
+++ b/include/boost/spirit/home/qi/detail/expectation_failure.hpp
@@ -11,6 +11,8 @@ file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 #pragma once
 #endif
 
+#include <boost/spirit/home/support/info.hpp>
+
 #include <stdexcept>
 
 namespace boost { namespace spirit { namespace qi {
--- a/include/boost/spirit/home/qi/detail/expectation_failure.hpp
+++ b/include/boost/spirit/home/qi/detail/expectation_failure.hpp
@@ -1,0 +1,31 @@
+/*=============================================================================
+Copyright (c) 2001-2011 Joel de Guzman
+Copyright (c) 2016      Frank Hein, maxence business consulting gmbh
+
+Distributed under the Boost Software License, Version 1.0. (See accompanying
+file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================*/
+#if !defined(SPIRIT_EXPECTATION_FAILURE_JULY_19_2016)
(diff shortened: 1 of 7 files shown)


include/boost/spirit/home/qi/detail/expectation_failure.hpp
```diff
@@ -0,0 +1,31 @@
+/*=============================================================================
+Copyright (c) 2001-2011 Joel de Guzman
+Copyright (c) 2016      Frank Hein, maxence business consulting gmbh
+
```
djowel: Not a biggie, but you probably shouldn't claim copyright on a file with contents lifted elsewhere. The basic rule here is that you add your copyright if there's significant contribution to the code in terms or actual content.

mxc-commons: Was to fast with copy & paste. Fixed. Sorry.

include/boost/spirit/home/qi/detail/expectation_failure.hpp
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+#endif
+
+#include <stdexcept>
+
+namespace boost { namespace spirit { namespace qi {
+    template <typename Iterator>
+    struct expectation_failure : std::runtime_error
+    {
+        expectation_failure(Iterator first_, Iterator last_, info const& what)
```
Lastique: Is an include for `info` missing?

mxc-commons: That was not really a question, was it? :) You are perfectly right.

I believe, I have to slow down things again for the sake of quality. I went into supersonice mode when I saw how fast #197 was merged. I apologize.

djowel: Can you run all tests before a PR, preferably with at least two compilers?

mxc-commons: After that issues showed up, I ran the build on the whole spirit test-suite. Before I did run on expect directive only in order to detect c++11-isms.

I found that the test suite does not build without errors on either toolset msvc-9.0, msvc-12.0 and msvc-14.0. The expect directive's tests pass on those three. The errors are not related to the expect directive and look quite serious. I seem not to be able to attach the output files here, so I will send them to the mailing list.

At the moment I only have the several MS toolsets installed. I understand that I will have to provide non-MS toolsets as well. I will do.

Lastique: I believe, Boost.Spirit doesn't have an includability test (a test that includes a single header to check that the header is self-sufficient; the test is run for all public headers), so missing includes can go unnoticed.

mxc-commons: The test suite does not build without compilation errors or failed tests with  gcc 6.1.0 (MinGW) either. This applies to the develop branch. I can provide an output file if it is of interest. gcc complains (almost) about the same things my several msvc toolsets do.

mxc-commons: @Lastique: Your believings are right.

mxc-commons: Can I close this PR?

PR #201, opened 2016-07-28
Fix for spirit_v2 qi/alternative test case

mxc-commons:
For me it looks like an issue involving the type traits `not_is_variant` and `not_is_optional`. `boost::optionals` which contain a `boost::variant`  (`boost::optional<boost::variant>`) are treated as variants, i.e. `not_is_variant` gives `mpl::false_`. 

This leads to `call_optional_or_variant` gets called with `mpl::false_` which seems to be incorrect for this type. 

Because `not_is_variant` and `not_is_optional` are not mutually exclusive in that context, I changed that to continue with `mpl::true` if the Attribute is an optional.

With that change the test case compiles. 

But I am actually not sure about the intention behind treating optionals as variants by the traits. If this design produces side effects qi relies on, my change might be not sufficient. 

The spirit_v2/qi test suite compiles successfully and all tests succeed now. So I assume, there are no side effects which stopped working.

I corrected a typo in a comment, too.

I did not look after the errors karma and lex produce. I am not into those right now.

### DIFF
--- a/include/boost/spirit/home/qi/detail/alternative_function.hpp
+++ b/include/boost/spirit/home/qi/detail/alternative_function.hpp
@@ -147,7 +147,6 @@ namespace boost { namespace spirit { namespace qi { namespace detail
         bool call_unused(Component const& component, mpl::true_) const
         {
             // return true if the parser succeeds
-            //BOOST_STATIC_ASSERT_MSG(spirit::traits::not_is_variant<Attribute, qi::domain>::value, "Optional");
             return call(component,
                 mpl::and_<
                     spirit::traits::not_is_variant<Attribute, qi::domain>,
--- a/include/boost/spirit/home/qi/detail/alternative_function.hpp
+++ b/include/boost/spirit/home/qi/detail/alternative_function.hpp
@@ -23,7 +23,7 @@ namespace boost { namespace spirit { namespace qi { namespace detail
     template <typename Variant, typename Expected>
     struct find_substitute
     {
-        // Get the typr from the variant that can be a substitute for Expected.
+        // Get the type from the variant that can be a substitute for Expected.
         // If none is found, just return Expected
 
         typedef Variant variant_type;
@@ -138,14 +138,16 @@ namespace boost { namespace spirit { namespace qi { namespace detail
         template <typename Component>
         bool call(Component const& component, mpl::false_) const
         {
+            // fix for alternative.cpp test case, FHE 2016-07-28
             return call_optional_or_variant(
-                component, spirit::traits::not_is_variant<Attribute, qi::domain>());
+                component, mpl::not_<spirit::traits::not_is_optional<Attribute, qi::domain>>());
         }
 
         template <typename Component>
         bool call_unused(Component const& component, mpl::true_) const
         {
             // return true if the parser succeeds
+            //BOOST_STATIC_ASSERT_MSG(spirit::traits::not_is_variant<Attribute, qi::domain>::value, "Optional");
             return call(component,
                 mpl::and_<
                     spirit::traits::not_is_variant<Attribute, qi::domain>,


djowel:
Seems that:

    template <typename T, typename Domain>
    struct not_is_variant<boost::optional<T>, Domain>
      : not_is_variant<T, Domain>
    {};

Is indeed suspicious. I don't recall what lead to that, but have you tried removing that instead? Does it lead to other errors?

mxc-commons:
Yes, that's where it comes from. But simply removing Holy Code? I did not even consider that yet. ;)



> Am 28.07.2016 um 11:57 schrieb Joel de Guzman <notifications@github.com>:
>
> Seems that:
>
> template <typename T, typename Domain>
> struct not_is_variant<boost::optional<T>, Domain>
>   : not_is_variant<T, Domain>
> {};
> Is indeed suspicious. I don't recall what lead to that, but have you tried removing that instead? Does it lead to other errors?
>
> —
>
>

mxc-commons:
But seriously: `not_is_variant` gets referenced from all over the place. Because `alternative_function` is the only place which causes a problem, I preferred that local fix.

djowel:
Same here, but I'd like to know the reasons why.

djowel:
Actually, thinking about it, although a local fix is the easy way out, it might not be the correct solution and we might be propagating something that is fundamentally wrong.

mxc-commons:
I do not agree. The error occurs in `alternative_function` which is used only by alternative.hpp. If test cases for the alternative operator are carefully designed and complete and all pass, we can be somewhat confident that this does not get propagated something wrong anywhere.

On the other hand, changing things globally by removing that special `optional<variant>` treatment is likely to break something somewhere. I do not know that, but I guess, the original author did implement that because it was required somewhere else. At that place that behaviour is needed and would be broken.

Btw, I actually do not like `lying names`. Something called `not_is_variant` should detect in particular that and nothing else.

Who originally authored that? We should ask that guy what the purpose was that he did it that way. Without knowing that I'd vote for the local fix. Impact can be much better and easier controlled.

djowel:
That author is me ;-) and I no longer remember the reason why. That is why it is pertinent to get to the bottom of it.

djowel:
I also do not like lying names. So touché ;-)

mxc-commons:
That was you? omg ;-)

djowel:
yeah! LOL. but don't rub it in :-) I'm already green in embarrassment ;P

djowel:
OK, anyway, since I'm the one guilty, It's my responsibility to find out why and try to get a fix.

mxc-commons:
Trying to give some consolation: 

I think there were never ever any user complaints about that. That means that nobody ever used an optional alternative.

mxc-commons:
Ok. :) So can I close this?

djowel:
I think the proper fix is to do something like (pseudo) is_variant || (is_optional && optional::enclosed_type is_variant)

djowel:
No, please don't. It's good to have it here. I'll look into it asap.

mxc-commons:
Ok. I will keep the fix active in my fork to have at least the spirit_v2/qi testsuite compile.

mxc-commons:
Your pseudo: It should return mpl::true_ if it is an optional with embedded variant and mpl::false_ if it is a pure variant (at the place I fixed it). Cause we are in the alternative operator we should have a variant anyway, shouldn't we? But it could be an embedded one.

mxc-commons:
I hope, I am not wrong with true and false. These not_ thingies twist my brain. :)

djowel:
One approach is to have the traits logic in the alternative implementation as an implementation detail, with a clear comment on the rationale. that way, we can have this fix sooner rather than later.

djowel:
that is what i am going too once i see it break when i "fix" the optional-variant thing.

jpcanepa:
I'm sorry to comment in this issue that is already closed, but I'm not versed enough in the implementation and I need to know if this is related to [the trac reported issue #12349](https://svn.boost.org/trac10/ticket/12349).

In there I've (recently, thought the issue is around a year old) added a very small grammar that fails to compile with gcc 6.2 on boost 1.63 (since boost 1.61). This conversation seems very related to what I'm seeing (a problem around alternative_function), but as I said, I'm not savvy enough as to make sure if the problems _are_ the same. 

I did take this change and applied it over a 1.63 version and the error changes (now failing in 

`../../../snps/include/boost/spirit/home/qi/detail/assign_to.hpp:44:16: error: no type named ‘value_type’ in ‘class snps_boost_1_63_0::variant<int, std::__cxx11::basic_string<char, std::char_t
raits<char>, std::allocator<char> > >’
         struct is_container_of_ranges`

I'm having serious trouble with an upcoming release due to this, so any help would be appreciated.

djowel:
Hmmm... Have you tried it? Does it work?

djowel:
pardon previous comment. I didn't read enough. OK, let's see what we can do. First I need to understand the test code. Let's continue discussion in the Trac ticket.

djowel:
OK, seems this is is already fixed in the latest devel which should already make it to the latest Boost release. See line 141
alternative_function.hpp

// fix for alternative.cpp test case, FHE 2016-07-28

PR #202, opened 2016-08-14
Lex: Fixed unique_ptr usage

Kojoley:
Lex is using `unique_ptr` from `movelib` and relying on a bug in it, described in [a ticket](https://svn.boost.org/trac/boost/ticket/11758).
Since that bug fixed in `movelib` Lex is uncompilable.

I have simplified unique_ptr usage by the following rules:
  - `delete ptr.release()` is equal to `ptr.reset()`
  - `delete a.release(); a = b;` is equal to `b.swap(a); b.reset();`

Now Lex test suite compiles and runs succesfully. VC++ 6 should be happy too despite the fact that Boost no longer supports it.

### DIFF
--- a/include/boost/spirit/home/support/detail/lexer/generator.hpp
+++ b/include/boost/spirit/home/support/detail/lexer/generator.hpp
@@ -495,21 +495,13 @@ protected:
                         delete *l_iter_;
                         *l_iter_ = overlap_.release ();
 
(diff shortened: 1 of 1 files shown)


hkaiser:
@Kojoley thanks!

mxc-commons:
@Kojoley May The Force Be With You! :)

PR #204, opened 2016-08-17
Qi.Tests: Wrap `r1 = r1 | eps` with `SPIRIT_NO_COMPILE_CHECK` macro

Kojoley:
Logic of the test is to prevent direct left recursion, but
currently because of it `b2` fails to compile `alternative.cpp`.

I have wrapped the case with `SPIRIT_NO_COMPILE_CHECK` like it
has been done in `char1.cpp` and `sequence.cpp` tests.

Actually the test fails at compilation only with std >= c++11,
but passes with std <= c++03.

PS: I do know that `alternative.cpp` does not compile for a
different reason too (because of problem described in #201)

### DIFF
--- a/test/qi/alternative.cpp
+++ b/test/qi/alternative.cpp
@@ -240,12 +240,15 @@ main()
     }
 
     {
+#ifdef SPIRIT_NO_COMPILE_CHECK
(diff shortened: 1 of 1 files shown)


djowel:
I'll merge #201 for now. I need to do a proper fix later.

PR #205, opened 2016-08-17
Karma.Tests: Added `regression_real_policy_sign.cpp` to test suite

Kojoley:
The test was added in 90d65d4be785050f65eeee03769cfb995c4adf5b,
but still is not used in the test suite.


djowel:
Kojoley, these are very nice! I really appreciate it!

PR #207, opened 2016-08-18
Qi: Fixed missing space in the nested template

Kojoley:
The bug was introduced in PR #201

### DIFF
--- a/include/boost/spirit/home/qi/detail/alternative_function.hpp
+++ b/include/boost/spirit/home/qi/detail/alternative_function.hpp
@@ -140,7 +140,7 @@ namespace boost { namespace spirit { namespace qi { namespace detail
         {
             // fix for alternative.cpp test case, FHE 2016-07-28
             return call_optional_or_variant(
-                component, mpl::not_<spirit::traits::not_is_optional<Attribute, qi::domain>>());
+                component, mpl::not_<spirit::traits::not_is_optional<Attribute, qi::domain> >());
         }
 
         template <typename Component>


Kojoley:
@djowel can I ask you to merge this PR?

djowel:
Right! :-) Thanks, Kojoley.

mxc-commons:
I am aware that the extra space was required because of a potential ambiguity with `operator >>` parsers couldn't cope with.

Are there still compilers requiring this? I'd like to add one of these to my test environment.

djowel:
It is required with all pre C++11 compilers (C++03)

mxc-commons:
msvc-9.0 is pre-11 is in my environment and has no problems. That's why I
asked.

Frank Hein
Geschäftsführer
Am Weißen Stein 1
41541 Dormagen
T  +49 2133 - 2599-10
F  +49 2133 - 2599-29
M +49 151 - 54742989
E frank.hein@maxence.de

djowel:
msvc-9.0 may be pre-c++11 but implements some C++11. Check if it has a flag with strict C++03 only. I doubt it. I don't know a strictly C++03 version of msvc. Maybe you are better off with GCC for testing C++03-isms.

mxc-commons:
msvc can not be set precisely to a particular c++ standard. Would it make sense to define a particular "minimum compliance platform"? Qi is some days old now. Would gcc-6.1.0 set to c++03 be ok as such a reference platform?

Kojoley:
GCC and Clang checks are made to explicitly fail on `>>` in templates if you are using `-std=c++03` or `-std=c++98`. This behavior allows you to use the latest compiler for development and do not be afraid of breaking something for compilers that not support latest C++ standard. Feel the difference from VS and undisaable M$ extensions.

PR #209, opened 2016-08-19
For illustration: Alternative to PR#208

mxc-commons:


mxc-commons:
Reject this. It's just to illustatrate what I think what would have been cleaner at The first. Not after a  decade of satisfied customers.

mxc-commons:
git probs.

PR #211, opened 2016-08-19
Qi: List operator not omitting second argument

mxc-commons:
For discussion: This is an example of the list operator not omitting the second argument's attribute.
For this one the original docu page's first part is true: `a % b` is equivalent to `a >> *(b >> a)`. 
The attribute section would need correction, because `b` would not longer get omitted.
See #208.

To retain the behaviour of the current list operator using the new one, you'd have to write `a % omit[b]`. An example file is provided to show the difference.

It feels strange that currently `a % char_('b')` is equivalent to `a % lit('b')`. Just a matter of taste?

### DIFF
--- a/example/qi/List.cpp
+++ b/example/qi/List.cpp
@@ -1,0 +1,67 @@
+/*=============================================================================
+Copyright (c) 2016 Frank Hein, maxence business consulting gmbh
+
+Distributed under the Boost Software License, Version 1.0. (See accompanying
+file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+==============================================================================*/
+#include <iostream>
+#include <map>
+
+#include <boost/spirit/home/qi.hpp>
+
+namespace qi = boost::spirit::qi;
+
+using iterator_type = std::string::const_iterator;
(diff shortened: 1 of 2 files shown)


djowel:
I think this one needs some more explanation on the use-cases.

mxc-commons:
With or without this PR there is no particular use case which was not expressible as `a >> *(b >> a)`.
Today `a % b` is not a full replacement, because the attribute of `b` is omitted always. 

Namespace qualified identifiers `ident::ident2::ident3` (including the `::`) for example could be expressed as `id % string("::")` with that PR. Without this PR this example would have to be `name >> *(string("::") >> name)`.

I proposed that because I'd find it more consistent if `a % char_('b')` was not equivalent to `a % lit('b')` and `a % b` was not equivalent to `a % omit[b]`.

Accepted PR #210 addresses the same topic without a change to the operator.

djowel:
Well, alas, it's a no-go if it will introduce a breaking change.

mxc-commons:
It would introduce a breaking change because users would have to switch to an explicit omit using the operator.

This PR comes 10 years late or so, I know. :)

PR #228, opened 2016-12-13
fix name hiding bug #12497

sehe:
::boost::long_long_type was hidden by ::boost::spirit::long_long_type silently making the specialization inactive

See also https://svn.boost.org/trac/boost/ticket/12497, thanks @K-ballo for the leg-work there, that made it much easier to spot.


sehe:
Oh darn. Only after reviewing the effective change against develop I see that @K-ballo happens to have fixed this a month ago, with only the spelling difference.

djowel:
Feel free to close this is it is indeed already fixed.

sehe:
What's more, <:: is a trigraph so for pre-c++11 compilers that could have been problematic

PR #231, opened 2017-02-10
silence unused parameter warning

bebuch:


djowel:
I prefer (Parser const& /*p*/) or simply (Parser const&)

bebuch:
The parameter is used, but only in debug mode via #ifdef a few lines below. ;-)

PR #233, opened 2017-06-19
Is this a typo in add_predicate.hpp which caused defects in x3::phrase_parse?

glacierx:
          , Context const& context, RContext& rcontext, Attribute& /*attr*/) const
        {
            Iterator i = first;
            return this->subject.parse(i, last, context, rcontext, unused);
        }

Guys,
Please take a look at : [add_predicate.hpp](https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/x3/operator/and_predicate.hpp)
The parameter _attr_ is commented and an unused is passed instead which will cause x3::parse_phrase left input attribute result unchanged. Is that a typo?


djowel:
Not sure. Could you provide a test case that exhibits the problem?

glacierx:
[emtpy_attribute.cpp](https://github.com/glacierx/spirit/blob/develop/test/x3/emtpy_attribute.cpp)

djowel:
Ah, I should have said "minimal" as in: http://boost-spirit.com/home/feedback-and-support/

glacierx:
Updated.
[empty_attribute.cpp](https://github.com/glacierx/spirit/blob/develop/test/x3/emtpy_attribute.cpp)

Or you can try the live version from [here](http://coliru.stacked-crooked.com/a/a5591ed449c88ca5)

djowel:
Please try to make it more minimal (see the link above).

glacierx:
[Live demo](http://coliru.stacked-crooked.com/a/f3142d4ae8e6b72f)

glacierx:
[This](https://wandbox.org/permlink/68QiVmHjjDr9yfCV) is more stable. However this case reproduce the empty attribute but doesn't crash the attribute printing.

glacierx:
[This](http://coliru.stacked-crooked.com/a/a5591ed449c88ca5) one crashed the attribute printing.

PR #234, opened 2017-06-19
Crash when print attribute in debug mode.

glacierx:
Please take a look. Thanks.


djowel:
OK, I'll need to look into this first; reverted the merge. Could you provide a test case that exhibits the problem?

djowel:
(reverting this for now: https://github.com/boostorg/spirit/pull/235)

glacierx:
[emtpy_attribute.cpp](https://github.com/glacierx/spirit/blob/develop/test/x3/emtpy_attribute.cpp)

djowel:
See my comment in the other PR

PR #236, opened 2017-07-26
Update PGI C++ compiler support

dkolsen-pgi:
Remove a PGI-specific workaround in the functor constructor.  The workaround was necessary several years ago to get class functor to compiler with PGI, but the workaround now causes compilation errors.


Lastique:
If the workaround is needed for older versions of the compiler, why not add a version check?

djowel:
Very good point, actually.

dkolsen-pgi:
The older versions of the PGI compiler that would benefit from the workaround are not actively being used by any customers that we (PGI) know of.  Adding a version check would add clutter to the code with no real benefit.

PR #237, opened 2017-08-01
X3: Support recursive rules that modify the context

togermer:
make_context() always pushed a new node to the context list when the context is modified (e.g., for "with", "skip", or "no_skip" directives). If such directives are used in recursive rules, this leads to an infinite loop of template instantiations, effectively limiting the use of recursive rules in x3.

To fix this, we first check if the tag for the new node is already contained in the context. If we find the tag, we remove the corresponding existing node from the context before pushing the new node with the requested tag.

In order to remove a context entry, we have to rebuild all context nodes up to this entry. We can reuse (i.e. link to) the tail of the old context after this entry.

In order to resolve life-time issues of newly created context nodes we added an aggregating implementation of struct context.

### DIFF
--- a/include/boost/spirit/home/x3/support/context.hpp
+++ b/include/boost/spirit/home/x3/support/context.hpp
@@ -16,57 +16,74 @@ namespace boost { namespace spirit { namespace x3
     template <typename ID, typename T, typename Next = unused_type>
     struct context
     {
-        context(T& val, Next const& next)
-            : val(val), next(next) {}
+        T& val;
+        Next next;
+    };
 
-        T& get(mpl::identity<ID>) const
-        {
-            return val;
-        }
-
-        template <typename ID_>
-        decltype(auto) get(ID_ id) const
-        {
-            return next.get(id);
-        }
-
+    template <typename ID, typename T, typename Next>
+    struct context<ID, T, Next const&>
+    {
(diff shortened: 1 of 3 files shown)


djowel:
Nice catch! Seems you have good familiarity with X3!

djowel:
I assume all tests pass?

tomilov:
@djowel offtopic: is there something new in C++17 to significally improve Spirit? Are you planning new version of Spirit?

togermer:
Yes, all X3 tests pass.

cppljevans:
What is the compile-time cost?  The paragraph in your 1st comment which starts with:

   To fix this, 

and the following paragraph worry me that compile times will suffer.  Are there any
benchmarks to measure compile times?

saki7:
This long-standing issue was effectively resolved sometime between 2017 and 2024. I actually encountered the recursive instantiation scenario while working on #788 and confirmed that the recursive rules are now correctly instantiated; without requiring any workarounds on user code.

I carefully examined current behavior and documented the specification on the comments.

https://github.com/boostorg/spirit/blob/328d940c372367a8d76a48ed956e870b56738ce9/include/boost/spirit/home/x3/core/skip_over.hpp#L74-L125

PR #239, opened 2017-08-21
simplify with directive

wanghan02:
1. when used as x3::with<ID>(x), where x is an lvalue, we inject an lvalue reference of x to the context.
2. when used as x3::with<ID>(x), where x is a const lvalue, we inject a const lvalue reference of x to the context.
3. when used as x3::with<ID>(x), where x is an rvalue, we move the rvalue to the member of with_directive and inject an lvalue reference of that member to the context.
4. No copy is applied; std::ref can still be used but not needed.
5. remove unused with_context

### DIFF
--- a/include/boost/spirit/home/x3/directive/with.hpp
+++ b/include/boost/spirit/home/x3/directive/with.hpp
@@ -15,21 +15,40 @@ namespace boost { namespace spirit { namespace x3
     ///////////////////////////////////////////////////////////////////////////
     // with directive injects a value into the context prior to parsing.
     ///////////////////////////////////////////////////////////////////////////
+    template <typename Subject, typename Derived, typename T>
+    struct with_value_holder
+      : unary_parser<Subject, Derived>
+    {
+        typedef unary_parser<Subject, Derived> base_type;
+        mutable T val;
+        with_value_holder(Subject const& subject, T&& val)
+          : base_type(subject)
+          , val(std::forward<T>(val)) {}
+    };
+    
+    template <typename Subject, typename Derived, typename T>
+    struct with_value_holder<Subject, Derived, T&>
+      : unary_parser<Subject, Derived>
+    {
+        typedef unary_parser<Subject, Derived> base_type;
+        T& val;
+        with_value_holder(Subject const& subject, T& val)
+          : base_type(subject)
+          , val(val) {}
+    };
+
     template <typename Subject, typename ID, typename T>
     struct with_directive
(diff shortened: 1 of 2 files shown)


djowel:
Looks good! Can you provide some test files?

wanghan02:
@djowel Test cases have been added to test/x3/with.cpp. 
correction for comment of rvalue: rvalue is moved to the member of with_directive but we actually inject a CONST lvalue reference of that member to the context because parse(...) is a const function. Still no copy is applied (better than before). But we cannot modify this injection (like before). Changes to the commits are not needed. If modification is needed, injecting an lvalue is the only way to do it.

wanghan02:
@djowel Test cases have been added to test/x3/with.cpp
**rvalue injection**: The 1st commit actually inject a const ref to the member variable because parse(...) is a const method. This behaves the same as before (but without copying), but dose not really make sense. I think the best way to use with directive is:

- Inject non-const lvalue: able to modify during parsing, the original lvalue is modified. 
- Inject rvalue: able to modify during parsing. 
- Inject const lvalue: not able to modify during parsing. 

So I brought back the with_value_holder and change the T const specialization to T& specialization, and makes rvalue injection modifiable. Test cases have been changed to adapt to the new requirement.

djowel:
wonderful!

include/boost/spirit/home/x3/directive/with.hpp
```diff
@@ -15,21 +15,40 @@ namespace boost { namespace spirit { namespace x3
     ///////////////////////////////////////////////////////////////////////////
     // with directive injects a value into the context prior to parsing.
     ///////////////////////////////////////////////////////////////////////////
+    template <typename Subject, typename Derived, typename T>
+    struct with_value_holder
+      : unary_parser<Subject, Derived>
+    {
+        typedef unary_parser<Subject, Derived> base_type;
+        mutable T val;
```
Kojoley: Could you explain why is it `mutable`?

1. The tests you had added do not cover it.
2. Allowing to mutate the value creates thread-safety risks.
3. It does not play well with `constexpr`, especially with Clang 6 and below.

wanghan02: The idea was rvalue injection should be mutable in const parser as well. Given parsers are immutable in the parse functions, rvalue injection would be immutable without `mutable` keyword. 

But I also agree with you on the issues it brings.  I don't have a solution at the moment.

PR #241, opened 2017-08-25
Add noexcept specs to x3::variant and forward_ast

dodheim:
This adds `noexcept` specifications to X3's `variant` and `forward_ast`. Missing specs on the types' move constructors was showing up in my profiling and I decided to fix up the types in whole.

I wasn't sure whether to use `noexcept` directly or `BOOST_NOEXCEPT`/`BOOST_NOEXCEPT_IF`; I couldn't find any precedent so opted for the latter.


djowel:
Wonderful!

PR #245, opened 2017-10-02
oss-fuzz: avoid signed integer overflow

dtardon:


djowel:
I'm not sure what to make of this. Please add a test case exhibiting the issue this PR is supposed to fix.

dtardon:
github doesn't let me attach a sample file, so I'll paste it... Compile with clang++/g++ with -fsanitize=undefined:
 
 #include <string>
 
 int main()
 {
     const std::string s = "0E-2147483648";
     std::string::const_iterator it = s.begin();
     qi::phrase_parse(it, s.end(), qi::double_, qi::space);
 }

dtardon:
And that didn't work too well either, so here it is again: https://paste.fedoraproject.org/paste/rs6RUHcwrXvY3oJeTgyHjQ .

djowel:
Are all the Qi tests still passing with this change?

dtardon:
@djowel - Yes, they do.

PR #256, opened 2017-10-31
Qi.Repository: Fix `keywords` compilation on c++11

Kojoley:
Though `result` specializations should be symmetrical to `operator()`, in this particular case author used 'hacky' way and I did not touch this, only fixed compilation error and done it in the same way as it is in other places e.g. https://github.com/boostorg/spirit/blob/9a8cded55355644c749d6511504b59c048d64de3/include/boost/spirit/home/karma/operator/alternative.hpp#L65-L69


Kojoley:
I think https://svn.boost.org/trac10/ticket/11493 could be closed now.

djowel:
Wonderful! Many thanks, Kojoley!

PR #259, opened 2017-11-03
Fix `subrule` compilation

Kojoley:
This fixes a long standing bugs persisted in `subrule` for years.
Those `%=` operator overloads are awful, but I do not know other solution (It should be possible to [tell proto to disable operators](http://www.boost.org/doc/libs/1_65_1/doc/html/proto/users_guide.html#boost_proto.users_guide.front_end.customizing_expressions_in_your_domain.inhibiting_overloads) but I do not know spirit's internals well enough for doing that).

Tested on VS2017, clang 5.0 c++03/1y, gcc 6 c++03/1y

### DIFF
--- a/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
+++ b/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
@@ -351,13 +351,20 @@ namespace boost { namespace spirit { namespace repository { namespace karma
 
         typedef mpl::vector<T1, T2> template_params;
 
-        // locals_type is a sequence of types to be used as local variables
+        // The subrule's locals_type: a sequence of types to be used as local variables
         typedef typename
             spirit::detail::extract_locals<template_params>::type
         locals_type;
 
+        // The subrule's encoding type
         typedef typename
-            spirit::detail::extract_sig<template_params>::type
+            spirit::detail::extract_encoding<template_params>::type
+        encoding_type;
+
+        // The subrule's signature
+        typedef typename
+            spirit::detail::extract_sig<template_params, encoding_type
+              , spirit::karma::domain>::type
         sig_type;
 
         // This is the subrule's attribute type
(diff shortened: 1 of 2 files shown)


Kojoley:
It took me several days to figure out the operators problem I really hope there is a simple method of disabling unused `proto` operators for `spirit` expressions.

djowel:
Wow, you really know Spirit! :-) I assume all tests are good with this patch?

BTW, would you like to be an active maintainer? Can't find your email in github, but feel free to email me at djowel-at-gmail-dot-com

Kojoley:
> Wow, you really know Spirit! :-)

Nah, I am still bad, but I make progress.

> I assume all tests are good with this patch?

Yes, with this patch `repository/tests` now finally fixed https://travis-ci.org/Kojoley/spirit/builds/296877440

> BTW, would you like to be an active maintainer?

It will make configuring CI stuff for Spirit repository easier for me, so yes, thanks for the offer!

Kojoley:
I have self-merged this because it works for me and cannot break already broken.

include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
```diff
@@ -447,6 +470,16 @@ namespace boost { namespace spirit { namespace repository { namespace karma
         }
 
         // non-const versions needed to suppress proto's %= kicking in
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
```
vtnerd: Do you know why the `operator%=(const&,&)` and `operator%=(const&,&&)` need to exist? Can't they be removed and use the `operator%=(const& const&)` version instead? It appears that the right hand side never needs to be modified.

Kojoley: You will get what was before the commit - proto's `operator%=(&&, &&)` will be used and it is not what we want.

PR #265, opened 2017-11-08
X3.Tests: Added regression test for 379413a5 fix

Kojoley:
Fix 379413a50c93539aa432fa9f9fc94fec1d6aa60d did not add a regression test.

For the first time I thought it fixes `vector<std::string, std::string>`, but it run smooth after reverting and it got me some minutes to come up with the test you see in this commit.

@djowel I hope it is what you meant in the commit message and I did not find other bonded regression.

### DIFF
--- a/test/x3/attr.cpp
+++ b/test/x3/attr.cpp
@@ -47,6 +47,11 @@ int main()
         std::string s;
         BOOST_TEST(test_attr("s", "s" >> attr(std::string("123")), s) &&
             s == "123");
+
+        s.clear();
+        using boost::spirit::x3::string;
+        BOOST_TEST(test_attr("123", string("123") >> attr(std::string("456")), s) &&
+            s == "123456");
     }
 
     return boost::report_errors();


djowel:
This is a bad case where a "fix" is not supported by a test case. I'm not sure how this got to this to begin with. Let's do it right. You are doing it right!

Kojoley:
Let's leave on hold this because I only now remembered that most of x3 test suite is actually an adapted version of Qi's test suite. I will return back to this after fixing remaining problems in Qi/Karma.

Kojoley:
I think 379413a50c93539aa432fa9f9fc94fec1d6aa60d was plainly wrong because it is a workaround for `attr` parser which advertises itself as `handles_container` but in reality it does not. The worst part about that change is that it introduced copying (#472 turned them into moves) in a lot of places where it is not needed at all.

The best way to sort this out is to make `attr` parser correctly handle containers and revert 379413a50c93539aa432fa9f9fc94fec1d6aa60d.

djowel:
> The best way to sort this out is to make `attr` parser correctly handle containers and revert [379413a](https://github.com/boostorg/spirit/commit/379413a50c93539aa432fa9f9fc94fec1d6aa60d).

Agreed. Will you be looking into this?

Kojoley:
> Will you be looking into this?

Yeah. The fix is pretty straightforward, we just need to turn this
https://github.com/boostorg/spirit/blob/38ec0cc2c1d0030d2527cab23e199073d382b45b/include/boost/spirit/home/x3/auxiliary/attr.hpp#L47
into `traits::move_to(traits::begin(value_), traits::end(value_), attr_);` when `value_` and `attr_` are both containers, I just need to find a good looking way to do it as we cannot use `if constexpr` :-(.

Kojoley:
Meh, that's actually uncovers a problem in `move_to`, it already should have handling this case with https://github.com/boostorg/spirit/blob/38ec0cc2c1d0030d2527cab23e199073d382b45b/include/boost/spirit/home/x3/support/traits/move_to.hpp#L82-L87
but for the same attributes an other overload is best match
https://github.com/boostorg/spirit/blob/38ec0cc2c1d0030d2527cab23e199073d382b45b/include/boost/spirit/home/x3/support/traits/move_to.hpp#L205-L210

PR #266, opened 2017-11-08
X3: Fix `unused_type` attribute case in `parse_into_container`

Kojoley:
Passing `unused_type` attribute to `parse_into_container_impl::call` produces a compilation error (clang: https://travis-ci.org/Kojoley/spirit/jobs/298783103 gcc: https://travis-ci.org/Kojoley/spirit/jobs/298783104).

The issue was introduced in 379413a50c93539aa432fa9f9fc94fec1d6aa60d.

While the fix restores x3 test suite (checked on msvc, gcc, clang) I am not quite sure it is the right solution. @djowel please, review the fix.

### DIFF
--- a/include/boost/spirit/home/x3/core/detail/parse_into_container.hpp
+++ b/include/boost/spirit/home/x3/core/detail/parse_into_container.hpp
@@ -245,6 +245,15 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
                 parser, first, last, context, rcontext, attr);
         }
 
+        template <typename Iterator>
+        static bool call(
+            Parser const& parser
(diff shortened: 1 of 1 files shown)


djowel:
Yes, you are doing it right. A "fix" without a test case is always suspicious.

Kojoley:
Without this fix current test suite does not pass (compile error in `confix` test, see logs above). You want me to come up with a new one?

djowel:
What "new one" do you have in mind?

djowel:
Anyway, I think this is correct as far as I can tell.

Kojoley:
That does not exist (or disabled) in current test suite and targets particular problem.

PR #279, opened 2017-11-15
Repository: subrule: Fixed partial ordering problem

Kojoley:
The problem is that `f(T&)` and `f(T&&)` overloads had the same priority
http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1164

Affected compilers are:
 - MSVC 9/10/11
 - GCC 4.7.2, 4.8.4 with `-std=c++1x`

### DIFF
--- a/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
+++ b/include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
@@ -458,90 +458,34 @@ namespace boost { namespace spirit { namespace repository { namespace karma
                 def_type(compile<spirit::karma::domain>(expr), name_)));
         }
 
-        template <typename Expr>
-        friend typename group_type_helper<Expr, true>::type
-        operator%=(subrule const& sr, Expr const& expr)
-        {
-            typedef group_type_helper<Expr, true> helper;
-            typedef typename helper::def_type def_type;
-            typedef typename helper::type result_type;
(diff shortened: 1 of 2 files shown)


include/boost/spirit/repository/home/karma/nonterminal/subrule.hpp
```diff
@@ -458,91 +458,35 @@ namespace boost { namespace spirit { namespace repository { namespace karma
...
-        friend typename group_type_helper<Expr, true>::type
-        operator%=(subrule const& sr, Expr const& expr)
-        {
-            typedef group_type_helper<Expr, true> helper;
-            typedef typename helper::def_type def_type;
-            typedef typename helper::type result_type;
-            return result_type(fusion::make_map<id_type>(
-                def_type(compile<spirit::karma::domain>(expr), sr.name_)));
-        }
+#define SUBRULE_MODULUS_ASSIGN_OPERATOR(lhs_ref, rhs_ref)                     \
```
vtnerd: Should this have a `BOOST_` prefix? It should be extremely rare to find this macro already declared elsewhere, but ...

Kojoley: It had `BOOST_SPIRIT_` but was shortened. Someone even could have and this one, so the actual solution is to add push/pop. I do not want to touch this more, because I have in mind an alternative fix and this overload madness should be removed.

PR #282, opened 2017-11-18
ofz#4303 avoid signed integer overflow

dtardon:
/usr/include/boost/spirit/home/qi/numeric/detail/real_impl.hpp:110:26: runtime error: signed integer overflow: -2147483647 - 19 cannot be represented in type 'int'

### DIFF
--- a/include/boost/spirit/home/qi/numeric/detail/real_impl.hpp
+++ b/include/boost/spirit/home/qi/numeric/detail/real_impl.hpp
@@ -108,6 +108,8 @@ namespace boost { namespace spirit { namespace traits
     inline bool
     scale(int exp, int frac, T& n, AccT acc_n)
     {
+        if (exp < std::numeric_limits<int>::min() + frac)
+            return false;
         return scale(exp - frac, n, acc_n);
     }
 


dtardon:
This has been found by oss-fuzz. Standalone reproducer: https://paste.fedoraproject.org/paste/Ab2L-UxCcjsJ1UDUOQ8jdw

Kojoley:
It looks like `x3` has the same problem or comment should be added https://github.com/boostorg/spirit/blob/1c8d9b6e25da80d414eb59709f41acb7d2c67e55/include/boost/spirit/home/x3/support/numeric_utils/extract_real.hpp#L62-L67

@djowel is the a reason why `qi` and `x3` does not share numeric parsing code? I find it ~80% identical.

kcc:
FYI: we are already fuzzing parts of boost on oss-fuzz: https://github.com/google/oss-fuzz/tree/master/projects/boost
Other parts, such as spirit, are welcome too.

djowel:
@Kojoley, the plan was to cleanup "support" and merge them. X3 "support" is cleaner and was meant to be the replacement for Qi support, somehow. I wanted to keep X3 clean.

djowel:
Oh, and before I forget, we need to add test cases for issues like these to make sure we won't do anything to break it in the future, esp. if we are going to merge X3 with Qi support. PRs such as these should be backed up by a test case that exhibits the behavior (of mis-behavior :)).

djowel:
> Oh, and before I forget, we need to add test cases for issues like these to make sure we won't do anything to break it in the future, esp. if we are going to merge X3 with Qi support. PRs such as these should be backed up by a test case that exhibits the behavior (of mis-behavior :)).

@dtardon can you do that ^ ? We need a test case that exhibits the problem this issue is fixing.

djowel:
I'm closing this one. A better way to address this is to not allow such big digits in the default real number policies for primitive floating point numbers. The current code (real policies) limits the number of digits to traits::max_digits10<T>::value. The new ignore_excess_digits policy should also have a limit as well.

See https://github.com/boostorg/spirit/issues/508

Issue #287, opened 2017-11-22
x3 cannot parse into associative containers anymore


octopus-prime:
Today i took a look at boost 1_66_0_b1 (because of boost beast).
But this broke my parser based on x3.

```
source/parser.cpp:48:53:   required from here
/home/mike/Downloads/boost_1_66_0_b1/boost_1_66_0/boost/spirit/home/x3/core/detail/parse_into_container.hpp:259:17: error: no matching function for call to ‘std::map<hessian::value_t, hessian::value_t>::insert(std::map<hessian::value_t, hessian::value_t>::iterator, std::map<hessian::value_t, hessian::value_t>::iterator, std::map<hessian::value_t, hessian::value_t>::iterator)’
                 attr.insert(attr.end(), rest.begin(), rest.end());
```

As you can see there is a `attr.insert(attr.end(), rest.begin(), rest.end());` call now.
This does not work, because no associative container (std::set, std::map, ...) provides this member function.

I think this bug is result of commit: https://github.com/boostorg/spirit/commit/379413a50c93539aa432fa9f9fc94fec1d6aa60d#diff-2b82341b4caa57a2a0c6bc9a616a1e1a

octopus-prime:
The good news is we have container_traits. So replacing

`attr.insert(attr.end(), rest.begin(), rest.end());`
by
`traits::append(attr, rest.begin(), rest.end());`

could help.

Kojoley:
Thanks for reporting this. I failed to find tests for associative containers in `x3` what is very disappointing. I am afraid the bug will not be fixed in 1.66 :disappointed:.

octopus-prime:
Ouch. The bugfix seems really easy. Tried it here successfully. And bugfix period for 1.66 ends at dec/06. So there is plenty of time...

Kojoley:
The master has not been updated since August and it contains not only bugfixes so we need to backport them from develop. Additionally `x3` does not run on boost regression and it is the second part of the problem.

octopus-prime:
That's disappointing. Can i help to improve our situation?

Kojoley:
We need to:
 - decide what should be backported  https://github.com/boostorg/spirit/compare/master...develop to fix the test suite (you can create a branch from master and cherry-pick merge commits from develop to it)
 - add tests
 - run the tests
 - write release notes

I can run tests on VS 2008, 2010, 2015, 2017 locally, and bunch of Clang/GCC on Travis.

Kojoley:
[...]

PR #289, opened 2017-11-23
Parse into associative containers

octopus-prime:
Bugfix and tests for issue #287 "x3 cannot parse into associative containers anymore"


octopus-prime:
The tests are useless if not executed...
We must add container_support.cpp to Jamfile again!

djowel:
Agreed. I wonder why it's not (or was removed)  in the Jamfile!

djowel:
Feel free to add it to the Jamfile. Seems you haven't yet in this PR?

octopus-prime:
@Kojoley says in #287 

> but container_support.cpp does not run because it is not present on Jamfile. I just tested and it passes (not on MSVC, because of the other problem)

So I think `container_support.cpp` is not included because of the problems with MSVC?!

djowel:
I don't think so. Anyway, If that's the case, then #ifdef it out for now. But, better yet, make it work on MSVC :-)

octopus-prime:
So, how to get a feedback from travis and appveyor?

PR #297, opened 2017-11-27
fix uint_parser<T(signed)> overflow problem

wanghan02:
According to spirit document,

  "All numeric parsers check for overflow conditions based on the type T the corresponding uint_parser<> has been instantiated with. If the parsed number overflows this type the parsing fails. Please be aware that the overflow check is not based on the type of the supplied attribute but solely depends on the template parameter T."

the valid input range of `x3::uint_parser<T>` should be `0` - `std::numeric_limits<T>::max()`. However, the current implementation takes values from `std::numeric_limits<T>::max()+1` - `std::numeric_limits<std::make_unsigned_t<T>>::max()` also as valid input when `T` is a signed integral.

Example:
  `x3::uint_parser<signed char>` should parse 0-127, but with current implementation 128-255 also works.

Test cases are added.

### DIFF
--- a/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
+++ b/include/boost/spirit/home/qi/numeric/detail/numeric_utils.hpp
@@ -504,35 +504,6 @@ namespace boost { namespace spirit { namespace qi { namespace detail
     };
 
 #undef SPIRIT_NUMERIC_INNER_LOOP
-
-    ///////////////////////////////////////////////////////////////////////////
-    // Cast an signed integer to an unsigned integer
-    ///////////////////////////////////////////////////////////////////////////
-    template <typename T,
-        bool force_unsigned
-            = mpl::and_<is_integral<T>, is_signed<T> >::value>
-    struct cast_unsigned;
-
-    template <typename T>
-    struct cast_unsigned<T, true>
-    {
-        typedef typename make_unsigned<T>::type unsigned_type;
-        typedef typename make_unsigned<T>::type& unsigned_type_ref;
-
-        inline static unsigned_type_ref call(T& n)
-        {
-            return unsigned_type_ref(n);
-        }
-    };
-
-    template <typename T>
-    struct cast_unsigned<T, false>
-    {
-        inline static T& call(T& n)
-        {
-            return n;
-        }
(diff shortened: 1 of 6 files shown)


djowel:
Is this for X3 or Qi? Or it seems both?

Kojoley:
Now for both.

djowel:
Great!

djowel:
looks good to me.

wanghan02:
@Kojoley Thanks!

Kojoley:
Thanks for the patch @wanghan02!

mwpowellhtx:
@djowel I'm not positive I've just bumped into this, or a similar, [kerfuffle](http://coliru.stacked-crooked.com/a/0c0f106c9952e2e1) as well in Qi. Base 10 parses negative values just fine, but has difficulty dealing with negative base 16 (hex) and base 8 (octal) integers. See [example code](http://coliru.stacked-crooked.com/a/0c0f106c9952e2e1) for illustration.

Kojoley:
> I'm not positive I've just bumped into this, or a similar

The problem in your example has nothing with the PR (it was about `uint_parser` while you are using `int_parser`), you can confirm that by testing on previous versions. What you are trying to achieve was never supported (int parsers explicitly expects minus sign in front of a number to threat it as a negative https://github.com/boostorg/spirit/blob/35c6a9a1bbf10004598f963604b8b881b347920d/include/boost/spirit/home/qi/numeric/numeric_utils.hpp#L120-L124). It is also a platform dependent thing, but should not be that hard for you to implement yourself.

Here is how I would done it (for two's complement; if I was needed to):
```cpp
template <typename T>
bool parse_int_hex(..., T& out)
{
    make_unsigned_t<T> value;
    if (!parse(..., value)) return false;
    constexpr auto max_positive = static_cast<make_unsigned_t<T>>(std::numeric_limits<T>::max());
    if (value > max_positive)
        out = -static_cast<T>(value - max_positive) + std::numeric_limits<T>::min();
    else
        out = static_cast<T>(value);
    return true;
}
```

mwpowellhtx:
@Kojoley Good to know, thank you for clarifying that. Apologies for the detour.

PR #304, opened 2017-12-01
Line pos iterator fix3

octopus-prime:
This is the fix for #305 "bug in line_pos_iterator".
Based on #40 "Fixing line_pos_iterator column numbering bug and get_current_line bug." from @tomoki


djowel:
looks good to me. is there a similar one in x3?

octopus-prime:
No, there is no similar one in x3.
And in my opinion there is no need for a similar one in x3.
`line_pos_iterator` is just an iterator adapter.
No dependencies to x3, qi, etc. But it can be used from all there, like `istream_iterator` too.

Kojoley:
> is there a similar one in x3?

Yes, it is.
https://github.com/boostorg/spirit/blob/79995f7e8fb00809745e99fd979d2cacc5a6a980/include/boost/spirit/home/x3/support/utility/error_reporting.hpp#L97-L190
https://github.com/boostorg/spirit/blob/79995f7e8fb00809745e99fd979d2cacc5a6a980/include/boost/spirit/home/x3/support/utility/testing.hpp#L242-L247

djowel:
when we get to clean up the code base. it would be good to unify the common parts of x3 and qi. fixing two similar but not 100% identical code is rather intractable.

Issue #250, opened 2017-10-16
optional_attribute is useless in x3/support/traits/attribute_category.hpp


cppljevans:
Unlike the other *_attribute tags, optional_attribute is not used anywhere.
Shouldn't there be:

    template <typename T>
    struct attribute_category<boost::optional<T>>
        : mpl::identity<optional_attribute> {};

Also, in move_to.hpp, shouldn't there be a move_to function for optional?

djowel:
Hmmmm.. seems you are right. If you have a patch we can test if the test-suite still passes.

cppljevans:
I've made changes to my local copy of:

https://github.com/cppljevans/spirit/tree/get_rhs

However, that branch has many other changes which
are not yet ready for a pull request.
Also, I'm no expert with git and I'm wary of
trying to isolate the changes into another
branch.

I've attached the output of git diff, if that's any help.

[git-diff.txt](https://github.com/boostorg/spirit/files/1391443/git-diff.txt)

cppljevans:
I've made changes to my local copy of:

https://github.com/cppljevans/spirit/tree/get_rhs

However, that branch has many other changes which
are not yet ready for a pull request.
Also, I'm no expert with git and I'm wary of
trying to isolate the changes into another
branch. 

Attached is a `git diff`.  Hope that helps.
[git-diff.txt](https://github.com/boostorg/spirit/files/1391476/git-diff.txt)

cppljevans:
On 10/16/2017 10:10 PM, Joel de Guzman wrote:
> Hmmmm.. seems you are right. If you have a patch we can test if the test-suite still passes.
> 
Unfortunately, I just tried the list.cpp test and it fails at the test:

https://github.com/cppljevans/spirit-experiments/blob/get_rhs/test/x3/list.cpp#L82

So, I wouldn't try this patch yet.
Investigating why the test fails.
Also, the debug.cpp test fails :(

-regards,
Larry

PR #307, opened 2017-12-01
optional_attribute

octopus-prime:
PR for issue #250 "optional_attribute is useless in x3/support/traits/attribute_category.hpp"

### DIFF
--- a/include/boost/spirit/home/x3/operator/detail/sequence.hpp
+++ b/include/boost/spirit/home/x3/operator/detail/sequence.hpp
@@ -359,6 +359,16 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
         return parse_sequence_plain(parser, first, last, context, rcontext, attr);
     }
 
+    template <typename Parser, typename Iterator, typename Context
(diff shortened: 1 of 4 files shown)


djowel:
Do you think we need a ticket for this with better description of the problem?

octopus-prime:
> Do you think we need a ticket for this with better description of the problem?

The ticket is here: #250
If there is a real problem?! I don't know. Tests are passing with and without patch...
So currently it seams to be just more code.
But later it might help to work on specialized optional stuff.
Or will work on optional stuff become harder because of the specialized code??!

> Unlike the other *_attribute tags, optional_attribute is not used anywhere.

Currently that's the only relevant point. There is `optional_attribute` without definition/usage...

> Hmmmm.. seems you are right. If you have a patch we can test if the test-suite still passes.

So here is the added patch and tests are passing. Currently that's all.
So the team should discuss about the value/sense of this PR. I am unsure...

djowel:
right, i got confused. sorry, i am blind :P. please go ahead.

PR #308, opened 2017-12-02
X3: Workaround VS2015 bugs

Kojoley:
These workarounds are enough to make X3 pass all the tests on VS2015 (19.00.24210).
I think they are not too awful. Otherwise VS2015 should be banned as unusable.


djowel:
Wow, that is very interesting! i won't even ask how you discovered/found such a fix. Please feel free to merge.

PR #310, opened 2017-12-04
sequential_or: Fixed random order execution of underlying parsers

Kojoley:
After two hours of fighting with the optimizer, I happily drew attention to
this little nifty bitwise operator in `any_if_ns` function.

Explanation of the bug: bitwise inclusive OR operator is not a sequence point
(per chapter §5.13 of C++14 standard), that's why at compiling the expression
`a() | b() | ... | z()` optimizer is allowed to rearrange the execution order
of the functions `a`, `b` ... `z`.

There is high chance that a lot of people were misguided by the bug and chose
not to use `sequential_or`.

I vaguely remember how about three years ago I thought that I am dumb and/or
the documentation is wrong when I tried to use the `sequential_or` but ended
with some workaround.

There are three possible fixes:
  - This one
  - Make the original `any_ns` and `any_if_ns` strict ordered
    (could potentially make permutations operator slower)
  - Break the `any_ns` and `any_if_ns` API and somehow generalize the code

### DIFF
--- a/include/boost/spirit/home/qi/operator/sequential_or.hpp
+++ b/include/boost/spirit/home/qi/operator/sequential_or.hpp
@@ -15,7 +15,7 @@
 #include <boost/spirit/home/qi/detail/pass_function.hpp>
 #include <boost/spirit/home/qi/detail/attributes.hpp>
 #include <boost/spirit/home/support/detail/what_function.hpp>
-#include <boost/spirit/home/support/algorithm/any_if_ns.hpp>
+#include <boost/spirit/home/support/algorithm/any_if_ns_so.hpp>
 #include <boost/spirit/home/support/handles_container.hpp>
 #include <boost/fusion/include/as_vector.hpp>
 #include <boost/fusion/include/for_each.hpp>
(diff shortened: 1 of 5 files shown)


djowel:
Very interesting insights! I honestly haven't thought about this until now! Do we have tests for this?

Kojoley:
There cannot be a test for this particular problem. For me it happened on `sequential_or` tests. The bug depends on compiler optimization strategy, inlining aggressiveness and phase of the moon .

Kojoley:
Or if someone know a test for `i++ + ++i` problem - I can adopt it for our case.

Example:
```cpp
#include <cstdio>

int main()
{
    return printf("Hello") | printf("World");
}
```

The example produces "WorldHello" for me on MSVC on any optimization strategy except `Od` and with any inlining strategy, but it prints "HelloWorld" on GCC and Clang with everything I have tried.

From my point of view if such a test exists - it requires some preprocessor macro or compiler argument. Actually this is a good idea for an in-compiler-fuzzer, but again, the bug will appear on already existing tests and does not require a dedicated test.

djowel:
OK, well, I haven't studied your solution yet, but could you give me a quick explanation on how the solution works?

Kojoley:
It is very simple, the crucial part is b1c74d04022ac532d7760175929448724456b92f.

Kojoley:
I cannot setup Appveyor on the repository until the problem is fixed. Should I remove `sequence_or` tests from the test suit on CI?

djowel:
I understand now. Sorry for the delay. Please go ahead.

PR #313, opened 2017-12-05
X3: Fixed include guard names collision

### LINKED ISSUE #257: Include guard names collision
I'm getting this error when trying to use karma in boost 1.65.1:
`'signbit': is not a member of 'boost::spirit::detail'`

It seems that `SPIRIT_SIGN_MAR_11_2009_0734PM` is already defined in:
`boost\spirit\home\x3\support\numeric_utils\sign.hpp`
which triggers the include guard of:
`boost\spirit\home\support\detail\sign.hpp`

I can work around it by using this code:
```C++
namespace boost {
	namespace spirit {
		namespace detail {
			using boost::spirit::x3::signbit;
		}
	}
}
```

Kojoley:
Closes #257 and https://svn.boost.org/trac10/ticket/12084

### DIFF
--- a/include/boost/spirit/home/x3/core/action.hpp
+++ b/include/boost/spirit/home/x3/core/action.hpp
@@ -4,8 +4,8 @@
     Distributed under the Boost Software License, Version 1.0. (See accompanying
     file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
 =============================================================================*/
-#if !defined(SPIRIT_ACTION_JANUARY_07_2007_1128AM)
-#define SPIRIT_ACTION_JANUARY_07_2007_1128AM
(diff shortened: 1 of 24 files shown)


octopus-prime:
Should we change the include guard names for qi too?

Example (include/boost/spirit/home/qi/action/action.hpp):
```cpp
#if !defined(SPIRIT_ACTION_JANUARY_07_2007_1128AM)
#define SPIRIT_ACTION_JANUARY_07_2007_1128AM
```
too
```cpp
#if !defined(BOOST_SPIRIT_QI_ACTION_JANUARY_07_2007_1128AM)
#define BOOST_SPIRIT_QI_ACTION_JANUARY_07_2007_1128AM
```
?

djowel:
It should be fine.

PR #316, opened 2017-12-06
Parse into range

octopus-prime:
See https://svn.boost.org/trac10/ticket/12928

```cpp
    std::string str("x123x"); 
    boost::iterator_range<boost::range_iterator<decltype(str)>::type> attr; 
    if( x3::parse( boost::begin(str), boost::end(str), x3::lit('x') >> x3::raw[+x3::digit] >> x3::lit('x'), attr ) ) { 
        std::cout<<"Match! attr = "<<attr<<std::endl; 
    } else { 
        std::cout<<"Not match!"<<std::endl; 
    }
```
produces compile error
```
/usr/local/include/boost/spirit/home/x3/support/traits/container_traits.hpp:176:15: error: no member named 'insert' in 'boost::iterator_range<__gnu_cxx::__normal_iterator<char *, std::__cxx11::basic_string<char> > >'

            c.insert(c.end(), first, last);

            ~ ^
```
The iterator_range is no container and must not go to container_traits.

### DIFF
--- a/include/boost/spirit/home/x3/operator/detail/sequence.hpp
+++ b/include/boost/spirit/home/x3/operator/detail/sequence.hpp
@@ -369,6 +369,16 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
         return parse_sequence_plain(parser, first, last, context, rcontext, attr);
     }
 
+    template <typename Parser, typename Iterator, typename Context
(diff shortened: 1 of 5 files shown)


djowel:
Absolutely!

wanghan02:
Great!

PR #318, opened 2017-12-07
Fix container traits for msvc

### LINKED ISSUE #298: X3: MSVC unordered containers `is_reservable<T>::value == false`
Until we have a good integration for boostorg/spirit in appveyor - this is what we get for now (via private appveyor account):

```
compile-c-c++ bin.v2\libs\spirit\test\x3\x3_container_support-p3.test\msvc-14.1\release\address-model-64\threadapi-win32\threading-multi\container_support.obj
container_support.cpp
libs\spirit\test\x3\container_support.cpp(181): error C2338: is_reservable problem
libs\spirit\test\x3\container_support.cpp(183): error C2338: is_reservable problem
libs\spirit\test\x3\container_support.cpp(185): error C2338: is_reservable problem
libs\spirit\test\x3\container_support.cpp(187): error C2338: is_reservable problem
```

https://ci.appveyor.com/project/octopus-prime/spirit

octopus-prime:
- Fixes #298 and is based on https://github.com/boostorg/spirit/commit/45902cbb4e949e7543bcf48d3a4ed1247cab4f6b provided by @Kojoley 
- Added container_support.cpp to x3 test suite again

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/container_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/container_traits.hpp
@@ -14,7 +14,6 @@
 #include <boost/detail/iterator.hpp>
 #include <boost/fusion/include/deque.hpp>
 #include <boost/tti/has_type.hpp>
-#include <boost/tti/has_member_function.hpp>
 #include <boost/mpl/identity.hpp>
 
 #include <vector>
@@ -35,7 +34,6 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
         BOOST_TTI_HAS_TYPE(size_type)
         BOOST_TTI_HAS_TYPE(reference)
         BOOST_TTI_HAS_TYPE(key_type)
-        BOOST_TTI_HAS_MEMBER_FUNCTION(reserve)
     }
 
     template <typename T>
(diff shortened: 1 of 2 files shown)


Kojoley:
That was unfriendly takeover. Good job.

octopus-prime:
Your fix is 8 days old but still no PR.
All the time the tests for container_support are not running.
I would be happy if you could provide your fix and of course we close this PR without merge afterwards.

It is not my intend to takeover your ideas! My only intend is to run all tests. Sorry!

Kojoley:
You are doing fine. Do not worry.

octopus-prime:
What should happen with this PR now?

Kojoley:
Trigger appveyor

Kojoley:
> size_t{} was because 64bit.

Imagine, I pass my custom `struct { void resize(std::int8_t); }` container.

> 0 is unsigned int and 32bit.

That's not right. Literal will be unsigned if you use `u` suffix and `int` is *at least* 16bit.

octopus-prime:
`signed int` you are right - and even worst...
And at least 16bit, ok. On my machine 32bit...

The main idea was to prove for existence of exact `reserve` signature in container class, wasn't it?

So here is signature from std::vector: `void reserve( size_type new_cap )`
From http://en.cppreference.com/w/cpp/container/vector/reserve

So what about `T::size_type`? That's the exact type...

And of course I think `0` is good enough - so `size_t{}` was too.

Kojoley:
That's what you propose:
```cpp
std::vector<int> v;
v.reserve(std::vector<int>::size_type{123});
```
I wonder, you are doing it this way? It is crazy.

octopus-prime:
LOL. So I used `size_t`. May be `std::size_t` would be better.
But `0` is fine. I imagine your custom `struct { void resize(std::int8_t); }` container.

PR #319, opened 2017-12-07
Qi: Permutations parser always initializes optionals

Kojoley:
Closes https://svn.boost.org/trac10/ticket/12473

Optional parser never fails so `attr = val;` always triggers and initializes
the `attr` value. The special case for optional attributes could be safely
removed as any underlying parser must already correctly handle optionals.


djowel:
Nods. Good catch!

PR #230, opened 2017-02-07
Weaken iterator requirements in qi::parse and qi::phrase_parse

yegord:
Fix for https://svn.boost.org/trac/boost/ticket/12817

PR #320, opened 2017-12-07
ForwardIterator -> ReadableIteratorConcept & ForwardTraversalConcept

wanghan02:
The concept of `ForwardIterator` is flawed because it mixed 2 sets of concepts (value access and traversal) into 1 package.

http://www.boost.org/doc/libs/1_65_1/libs/iterator/doc/new-iter-concepts.html

It requires `value_type (const)&` as return type when dereference is applied, which is not mandatory in spirit parsing. A return type which is convertible to value_type is good enough. `ReadableIteratorConcept` and `ForwardTraversalConcept` should be what we need for the iterator check.

For example, the iterator of the range returned by `boost::adaptors::transform(std::string, func)` is normally not a `ForwardIterator`. But it fulfills `ReadableIteratorConcept` and `ForwardTraversalConcept` and should be able to be parsed by spirit.

Test cases are added for both qi and x3. 

There is also a ticket for this but the solution suggested is not correct. 
https://svn.boost.org/trac10/ticket/12817

### DIFF
--- a/include/boost/spirit/home/qi/parse.hpp
+++ b/include/boost/spirit/home/qi/parse.hpp
@@ -15,7 +15,7 @@
 #include <boost/spirit/home/support/context.hpp>
 #include <boost/spirit/home/support/nonterminal/locals.hpp>
 #include <boost/spirit/home/qi/detail/parse.hpp>
-#include <boost/concept_check.hpp>
+#include <boost/iterator/iterator_concepts.hpp>
 
 namespace boost { namespace spirit { namespace qi
 {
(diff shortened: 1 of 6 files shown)


wanghan02:
@Kojoley pull request #230 was not a correct solution (same as in the ticket).

djowel:
Hmmm, seems my previous comment did not get through? Or is this a different PR?

Kojoley:
Who knows.

djowel:
OK... anyway, what I wrote was: I like the new iterator traits, but it is not really "new". It was written in 2003 and so far, I don't see it assimilated by the standards (C++11 and onwards). Correct me if I am wrong. So, unless there's real activity towards that, this falls into the category of "don't fix it if it ain't broken".

Kojoley:
The PR is not about simply switching to "new" iterator concepts, but to make spirit less paranoid and allow the things like `boost::transformed_range`. Currently it produces the compilation error because of asserts and will work fine if you remove them.

djowel:
I see that now. OK, we need this alright. It fixes something that's broken. I'll merge this one.

djowel:
And thanks @wanghan02 and @Kojoley

PR #323, opened 2017-12-14
Classic: Fix compilation with BOOST_DISABLE_THREADS defined

Kojoley:
Closes https://svn.boost.org/trac10/ticket/12639

Spirit should not include any Boost.Thread header until user explicitly
requested threadsafe support with BOOST_SPIRIT_THREADSAFE defined.


djowel:
Agreed! Indeed.

PR #324, opened 2017-12-14
Lex: Fixed C2001 'newline in constant' warning in generated lexers

Kojoley:
Closes https://svn.boost.org/trac10/ticket/11540.

Actually this should have been caught by `regression_matlib_generate_switch`
test, but it is not complex enough to cover all cases. We do not have the
coverage reports yet, but the fix is trivial so I think it is ok.


hkaiser:
Could you explain what you had in mind with this patch, please? From what I see you're now comparing against a multi-character literal - is that legal in the first place?

hkaiser:
Sorry, I take that back - I see now what's you're doing.

PR #327, opened 2017-12-15
Auto attribute propagation failed when semantic action present

octopus-prime:
See https://svn.boost.org/trac10/ticket/5713

Added suggested test. Passed.

### DIFF
--- a/test/qi/rule2.cpp
+++ b/test/qi/rule2.cpp
@@ -157,6 +157,32 @@ main()
         BOOST_TEST(s == "abcdef");
     }
 
+    {
+        typedef boost::fusion::vector<int, int> V;
+        typedef rule<char const*, V(), space_type> R;
+        V val;
+        R r;
+        r %= int_ >> int_;
+        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
+    }
+
+    {
+        typedef boost::fusion::vector<int, int> V;
+        typedef rule<char const*, V(), space_type> R;
+        V val;
+        R r;
+        r %= int_ >> int_[_1];
+        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
+    }
+
+    {
+        typedef boost::fusion::vector<int, int> V;
+        typedef rule<char const*, V(), space_type> R;
+        V val;
+        R r;
+        r %= (int_ >> int_)[_1];
+        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
+    }
+
     return boost::report_errors();
 }
-
--- a/test/qi/rule2.cpp
+++ b/test/qi/rule2.cpp
@@ -162,24 +162,6 @@ main()
         typedef rule<char const*, V(), space_type> R;
         V val;
         R r;
-        r %= int_ >> int_;
-        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
-    }
-
-    {
-        typedef boost::fusion::vector<int, int> V;
-        typedef rule<char const*, V(), space_type> R;
-        V val;
-        R r;
-        r %= int_ >> int_[_1];
-        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
-    }
-
-    {
-        typedef boost::fusion::vector<int, int> V;
-        typedef rule<char const*, V(), space_type> R;
-        V val;
-        R r;
         r %= (int_ >> int_)[_1];
         BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
     }


test/qi/rule2.cpp
```diff
@@ -157,6 +157,32 @@ main()
         BOOST_TEST(s == "abcdef");
     }
 
+    {
+        typedef boost::fusion::vector<int, int> V;
+        typedef rule<char const*, V(), space_type> R;
+        V val;
+        R r;
+        r %= int_ >> int_;
```
Kojoley: This is redundant https://github.com/boostorg/spirit/blob/95e1432c0c81bb107283ffaedc1d4c15f57ee747/test/qi/sequence.cpp#L120

octopus-prime: So, what you suggest for this PR? Close?!

octopus-prime:
Or should there only 1 test (instead of 3):

```
        typedef boost::fusion::vector<int, int> V;
        typedef rule<char const*, V(), space_type> R;
        V val;
        R r;
        r %= (int_ >> int_)[_1];
        BOOST_TEST((test_attr("1 5", r, val, space) && val == V(1, 5)));
```

Since this failed in 1.47.0...

test/qi/rule2.cpp
Kojoley: Remove the redundant test case.

Kojoley:
https://svn.boost.org/trac10/ticket/5713 was fixed with https://github.com/boostorg/spirit/commit/f2b627ed9985bb09518d4e089a31575a32470b01, test case added in https://github.com/boostorg/spirit/commit/4727fc8f495809a0b6f23847b1309cf35a5c3620.

octopus-prime:
Are you sure that
https://github.com/boostorg/spirit/commit/95e1432c0c81bb107283ffaedc1d4c15f57ee747 and/or
https://github.com/boostorg/spirit/commit/4727fc8f495809a0b6f23847b1309cf35a5c3620 contain the same test?!

They look so different...

octopus-prime:
Still unsure what to do here. Just closing this PR?

octopus-prime:
Undoing https://github.com/boostorg/spirit/commit/f2b627ed9985bb09518d4e089a31575a32470b01 has no impact on the test cases provided here. https://github.com/boostorg/spirit/commit/f2b627ed9985bb09518d4e089a31575a32470b01 might be NOT the fix and https://github.com/boostorg/spirit/commit/4727fc8f495809a0b6f23847b1309cf35a5c3620 not the right test cases.

Kojoley:
That's because you are trying to cheat and you don't understand the problem.
The test case you are adding only catches the bug in c++03, because `boost::fusion::vector<int,int>` and `boost::fusion::vector2<int,int>` are the same types in c++11.

```cpp
#include <boost/fusion/container/vector.hpp>
#include <boost/type_traits/is_same.hpp>
#include <boost/mpl/assert.hpp>

int main()
{
    typedef boost::fusion::vector<int,int> a;
    typedef boost::fusion::vector2<int,int> b;
    BOOST_MPL_ASSERT(( boost::is_same<a, b> ));
    return 0;
}
```

The right test case should be something like this (if I understand the problem correctly):
```cpp
#include <boost/detail/lightweight_test.hpp>

namespace custom_parser 
{ 
    BOOST_SPIRIT_TERMINAL(bug)
} 

namespace boost { namespace spirit 
{ 
    template <>
    struct use_terminal<qi::domain, custom_parser::tag::bug> 
      : mpl::true_ 
    {}; 
}}

namespace custom_parser 
{ 
    struct bug_helper
    {
        bug_helper() {}
        bug_helper(int) {}
        operator int() const { return 123; }
    };

    struct bug_parser 
      : boost::spirit::qi::primitive_parser<bug_parser>
    {
        template <typename Context, typename Iterator>
        struct attribute
        {
            typedef bug_helper type;
        };

          , typename Skipper, typename Attribute>
          , Context&, Skipper const& skipper, Attribute& attr) const
        {
            boost::spirit::qi::skip_over(first, last, skipper);
            boost::spirit::traits::assign_to(bug_helper(), attr);
            return true;
        }

        template <typename Context>
        boost::spirit::info what(Context&) const
        {
            return boost::spirit::info("bug");
        }
    };
}

namespace boost { namespace spirit { namespace qi
{
    template <typename Modifiers>
    struct make_primitive<custom_parser::tag::bug, Modifiers>
    {
        typedef custom_parser::bug_parser result_type;

        result_type operator()(unused_type, unused_type) const
        {
            return result_type();
        }
    };
}}}

int main()
{
    using namespace boost::spirit::qi;
    rule<char const*, int() > r;
    r %= custom_parser::bug[_1];
    int val;
    char const* const it = "";
    BOOST_TEST((parse(it, it, r, val)));
    BOOST_TEST((val == 123));
}
```
It is similar check to what have been done in `regression_binary_action.cpp`, just without a custom parser (because it relies on binary parser internal structure).
If @hkaiser agree with me, I think `regression_binary_action.cpp` need to be rewritten with a custom parser and something similar added to x3 tests suite.

octopus-prime:
https://svn.boost.org/trac10/ticket/5713 is closed - so is this PR.

Kojoley:
A shorter test case:
```cpp
#include <boost/detail/lightweight_test.hpp>

struct custom_int
{
    custom_int() {}
    custom_int(int) {}
    operator int() const { return 123; }
};


int main()
{
    using namespace boost::spirit::qi;
    rule<char const*, int() > r;
    r %= attr(custom_int())[_1];
    int val;
    char const* const it = "";
    BOOST_TEST((parse(it, it, r, val)));
    BOOST_TEST((val == 123));
}
```

PR #330, opened 2017-12-17
X3: Reenable fusion::map support

Kojoley:
`build_container<char>` specialization was added in 84c0c075 (#5)


djowel:
Oh cool! Hmmm "re-enable"? You mean it was disabled at one point? Perhaps unintentionally?

Kojoley:
Compilation error because of missing include (but that was fixable at user side) and commented out `save_to_assoc_attr<variant<T...> >` specialization (turned off support for sequences where key is a variant).

djowel:
Right. I guess that happened with the incremental port from Qi. Thanks!

PR #336, opened 2017-12-19
Classic: Increment scanner through iterator policy

Kojoley:
Closes https://svn.boost.org/trac10/ticket/7371

The trac ticket author was right and left unnoticed for 5 years.


djowel:
👍

Kojoley:
@martinsjogren No, it is not an expected behavior. Thanks for discovering and reporting the regression.
@djowel I think cfe94349e220516b1c1a689273ee408790c815d3 (`fix-for-pr336` branch) should fix the problem, but it will be fine to simply revert the PR.

djowel:
OK. I'm fine with either approach. Looks like a regression alright.

djowel:
Yes it will be merged.

PR #337, opened 2017-12-19
X3: Fixed iterator move_to to single item sequence

Kojoley:
This affects lazy attribute constructing parsers, which exposes not an actual
attribute, but a pair of iterators (like `string` or `attr("...")`).


djowel:
Nice. I needed that.

PR #340, opened 2017-12-22
X3: Dereference a single item view instead of unwrapping sequence

Kojoley:
Most of discussion happened in https://github.com/boostorg/spirit/pull/219#issuecomment-352905393.
Brief overall: Currently we do not know what a8e391b was fixing (and it seems partly redundant), but it introduced an unwanted behavior of passing `fusion::iterator_range` to attribute-generating parsers.

This PR reimplements a spirit of a8e391b commit. It tries to pass through as many attributes as
possible. For example: a single item sequences like `fusion::vector<T>` will be passed through,
while fusion iterators, fusion iterator range or view of size one will be dereferenced before passing
to underlying parsers of sequence parser.

TODO:
  - Tests? It is not that easy to test, but I can make a custom parser and check attribute type in it.
  - Should `tuple_traits.hpp` be renamed? The name already does not reflect the content.

### DIFF
--- a/include/boost/spirit/home/x3/operator/detail/sequence.hpp
+++ b/include/boost/spirit/home/x3/operator/detail/sequence.hpp
@@ -103,30 +103,30 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
         }
     };
 
-    template <typename Parser, typename Attribute, bool pass_through>
+    template <typename Parser, typename Attribute>
     struct pass_sequence_attribute_used :
-        mpl::if_c<
-            (!pass_through && traits::is_size_one_sequence<Attribute>::value)
+        mpl::if_<
+            traits::is_size_one_sequence<Attribute>
           , pass_sequence_attribute_front<Attribute>
           , pass_through_sequence_attribute<Attribute>>::type {};
 
-    template <typename Parser, typename Attribute, bool pass_through = false, typename Enable = void>
+    template <typename Parser, typename Attribute, typename Enable = void>
     struct pass_sequence_attribute :
         mpl::if_<
             fusion::result_of::empty<Attribute>
           , pass_sequence_attribute_unused
-          , pass_sequence_attribute_used<Parser, Attribute, pass_through>>::type {};
+          , pass_sequence_attribute_used<Parser, Attribute>>::type {};
 
-    template <typename L, typename R, typename Attribute, bool pass_through>
-    struct pass_sequence_attribute<sequence<L, R>, Attribute, pass_through>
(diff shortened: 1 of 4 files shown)


djowel:
- Looks good to me
- I like the idea of a custom parser test
- Any suggestion on renaming tuple_traits.hpp? It still looks good to me.

Kojoley:
> Any suggestion on renaming tuple_traits.hpp? It still looks good to me.

It does not contain a single word 'tuple' except in a include guard. It should have been named `sequence_traits`, but now, with `is_size_one_view` added, I do not know.

djowel:
> Any suggestion on renaming tuple_traits.hpp? It still looks good to me.
> 
> It does not contain a single word 'tuple' except in a include guard. It should have been named sequence_traits, but now, with is_size_one_view added, I do not know.

Ah, it's the name "tuple". It's just that it is an attribute category name in X3 for "tuple" like types (tuple_attribute). Let's keep it as it is.

Kojoley:
I am too tired of complexity. It is already looks pretty scary. I wish C++ has Python generators.

djowel:
> I am too tired of complexity. It is already looks pretty scary. I wish C++ has Python generators.

Yeah, I wish the code can be simplified. I too am tired of such complexity. It's hard to reason out what the semantics of the code is by looking at the code. Perhaps there's a way to conceptualize the semantics with some smart meta-constructs.

Kojoley:
It can be simplified by removing the test :D
Currently it does not check a lot of cases, adding generator for permutations will require more code. I do not know how to simplify this except placing hundred hardcoded types.

djowel:
Let's just have this as it is now. Is there any other reason that's keeping you from pushing the code?

Kojoley:
> Is there any other reason that's keeping you from pushing the code?

The solution looks solid for me. The test catches both problems before and after a8e391b.
So, I think it is ready for merge.

djowel:
👍 Feel free to merge, @Kojoley

Kojoley:
I am afraid of adding anything to `namespace std`. Because of `BOOST_TEST_EQ` I was needed to add a stream operator for `std::pair`. How do you think, may be just replace it with `BOOST_TEST(a == b)`?

djowel:
Fair point. Yeah, it's best to avoid adding to namespace std (best practice, etc.). There should be a couple of ways, BOOST_TEST(a == b) being one of them.

Kojoley:
Huh, made the test twice shorter.

djowel:
OK, LGTM 👍

PR #345, opened 2017-12-29
Inheriting std::iterator is deprecated in c++17

DanielaE:
Boost's iterator.hpp is deprecated, too. It does nothing but pulling std::iterator into namespace boost and including standard headers 'iterator' and 'cstddef'. Therefore get rid of all of that and replace inheritance by lifting std::iterator's members into the derived class.

### DIFF
--- a/include/boost/spirit/home/classic/iterator/multi_pass.hpp
+++ b/include/boost/spirit/home/classic/iterator/multi_pass.hpp
@@ -16,7 +16,6 @@
 #include <algorithm>    // for std::swap
 #include <exception>    // for std::exception
 #include <boost/limits.hpp>
-#include <boost/iterator.hpp>
 
 #include <boost/spirit/home/classic/namespace.hpp>
 #include <boost/spirit/home/classic/core/assert.hpp> // for BOOST_SPIRIT_ASSERT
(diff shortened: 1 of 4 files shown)


include/boost/spirit/home/classic/iterator/multi_pass.hpp
```diff
@@ -770,14 +769,13 @@ struct iterator_base_creator
...
 
-    typedef boost::iterator
-    <
-        std::forward_iterator_tag,
-        typename input_t::value_type,
-        typename input_t::difference_type,
-        typename input_t::pointer,
-        typename input_t::reference
-    > type;
+    struct type {
```
Kojoley: It looks like the comment above is obsolete because of https://github.com/boostorg/core/commit/bca30b2ec7f41e6ece057b46d1358f0506c904d3.

DanielaE: No, it's not. Because of the change by that commit, inheriting from boost::iterator or referring to its members is just the same as inheriting from or using std::iterator. This is deemed deprecated now since C++17 and the latest version of MSVC spits out nasty warnings because of that. On top of that, Boost's version of iterator.hpp is deprecated as well.

Kojoley:
```
 // Meta-function to generate a std::iterator<> base class for multi_pass. This
 //  is used mainly to improve conformance of compilers not supporting PTS
 //  and thus relying on inheritance to recognize an iterator.
```
I do not think this part makes any sense after the change.

```
 // We are using boost::iterator<> because it offers an automatic workaround
 //  for broken std::iterator<> implementations.
```
And this since https://github.com/boostorg/core/commit/bca30b2ec7f41e6ece057b46d1358f0506c904d3.

DanielaE: Ah, *that's* your objection. Nevermind - I'll change that asap.

djowel:
👍 LGTM

Kojoley:
Thanks!

PR #347, opened 2018-01-03
Boost spirit instantiate

cppljevans:
* Problem:

    Separate compilation of grammar rules using the
    `BOOST_SPIRIT_INSTANTIATE` macro, under some
    circumstances( maybe when attribute is *not* of type
    unused_type *and* there's no semantic action?) leads
    to link error.
    
    The problem occurs because the
    `BOOST_SPIRIT_INSTANTIATE` macro generates a
    `parse_rule` instantiation requiring a specific type,
    `rule_type::attribute_type`, for the rule's attribute
    argument. In contrast, the `rule<ID,Attribute>::parse`
    function takes an arugment `ActualAttribute& attr` and
    passes that to the `parse_rule` function.  When
    `Attribute != ActualAttribute`, the link error occurs.
    
* Interest in solution.
    * [Exagon.linking-error-when](http://boost.2283326.n4.nabble.com/Linking-error-when-changing-to-tp4689820.html)
    * [OlafPeter.linkerr-separate-tu](https://stackoverflow.com/questions/43791079/x3-linker-error-with-separate-tu)
    * [Halvorsen.instant-linkerr](http://boost.2283326.n4.nabble.com/X3-declare-define-instantiate-linker-error-tp4694282.html)

* Method of solution.

  In the current `spirit` method, the `ActualAttribute&
  attr` that is passed to the `parse_rule` function that
  is called by `rule<ID,Attribute>::parse` is transformed,
  after a few intervening function calls, into `Attribute&
  _attr` by code at the beginning of the:
  
    > `detail::rule_parser<...>::call_rule_definition` 
  
  function.  This `attribute transform code` extends
  [from here](https://github.com/boostorg/spirit/blob/master/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L306)
  [to here](https://github.com/boostorg/spirit/blob/master/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L347), 
  execept for the `parsing/debugging code` actually doing the
  parsing.  This `parsing/debugging code` extends
  [from here](https://github.com/boostorg/spirit/blob/master/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L319)
  [to here](https://github.com/boostorg/spirit/blob/master/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L341)
  and separates the `attribute pre-transform code` from
  the `attribute post-transform code`.
    
  The link problem solution requires doing the attribute
  transformation in different places.
  To avoid code duplicaton, the combination of the `attribute transform code`
  and the `parsing/debugging code` mentioned in the
  previous paragraph has been replaced by the
  `rule_attr_transform_f` function in `detail/rule.hpp`.
  The `parsing` part of the `parsing/debugging code`
  mentioned in the previous paragraph is passed to this
  function by the `parser` argument.
  
  The link problem is solved by calling the
  `rule_attr_transform_f` in `rule<...>::parse` instead of
  calling `parse_rule` and removing it from
  `detail::rule_parser<...>::call_rule_definition`.
  However, without this transformation called from within,
  indirectly, `rule_definition<...>::parse`, using an
  instance of `rule_definition<...>` on the rhs of a
  grammar expression would result in no attribute
  transformation, which of course would be invalid.  But
  the transformation cannot be put in
  `rule_definition<...>::parse` because that would
  duplicate the transformation in `rule<...>::parse` when
  it calls, indirectly, the `rule_definition<...>::parse`
  via the `parse_rule` generated by the
  `BOOST_SPIRIT_DEFINE` macro.  The solution is to have
  this generated `parse_rule` call instead the new
  function `rule_definition<...>::parse_no_xform` which
  does no attribute transform.
  
* test code:
    * [test3/linker_error_example](https://github.com/cppljevans/spirit-experiments/blob/get_rhs/test/x3/linker_error_example/)

### DIFF
--- a/test/x3/linker_error_test/ast.hpp
+++ b/test/x3/linker_error_test/ast.hpp
@@ -1,0 +1,76 @@
+//
+// Created by lukas on 11.11.16.
+//
+
+#ifndef LINKER_ERROR_EXAMPLE_AST_HPP_HPP
+#define LINKER_ERROR_EXAMPLE_AST_HPP_HPP
+
+#include <boost/fusion/adapted/struct.hpp>
+//#define BOOST_SPIRIT_X3_DEBUG
+//#define PRAGMA_X3_DEBUG
+#ifdef PRAGMA_X3_DEBUG
+#ifdef BOOST_SPIRIT_X3_DEBUG
+    #pragma message "yesdef(BOOST_SPIRIT_X3_DEBUG)"
+#else
+    #pragma message "notdef(BOOST_SPIRIT_X3_DEBUG)"
+#endif
+#endif//PRAGMA_X3_DEBUG
+#include <boost/spirit/home/x3.hpp>
+#include <boost/spirit/home/x3/support/ast/variant.hpp>
+#define VARIABLE_DEC_OPTIONAL
+  //The reason for this macro is apparent
+  //in example_def.hpp
+//#define PRAGMA_VARIABLE_OPTIONAL
+#ifdef PRAGMA_VARIABLE_OPTIONAL
+#ifdef VARIABLE_DEC_OPTIONAL
+    #pragma message "yesdef(VARIABLE_DEC_OPTIONAL)"
+    #include <boost/optional/optional_io.hpp>//in response to compile-time err msg.
+#else
+    #pragma message "notdef(VARIABLE_DEC_OPTIONAL)"
+#endif
+#endif//PRAGMA_VARIABLE_OPTIONAL
+
+#include <vector>
+#include <string>
+
+namespace client { namespace ast {
+
+namespace x3 = boost::spirit::x3;
+
+struct LambdaType;
+struct ClassType;
+struct nil{};
+
+typedef x3::variant <
+    nil,
+    x3::forward_ast<LambdaType>,
+    x3::forward_ast<ClassType>
+> Type;
+
+struct LambdaType {
+    std::vector<Type> parameters_;
+    Type return_type_;
+};
+
+struct ClassType {
+    std::vector<std::string> name_;
+    std::vector<Type> template_args_;
+};
+
+struct VariableDec {
+    std::string _name;
+  #ifdef VARIABLE_DEC_OPTIONAL
+      boost::optional<Type> 
+  #else
+      Type
+  #endif
+    _type;
+};
+
+}}
+
+BOOST_FUSION_ADAPT_STRUCT(client::ast::LambdaType, parameters_, return_type_)
+BOOST_FUSION_ADAPT_STRUCT(client::ast::ClassType, name_, template_args_)
+BOOST_FUSION_ADAPT_STRUCT(client::ast::VariableDec, _name, _type)
+
+#endif //LINKER_ERROR_EXAMPLE_AST_HPP_HPP
--- a/test/x3/linker_error_test/config.hpp
+++ b/test/x3/linker_error_test/config.hpp
@@ -1,0 +1,18 @@
+#ifndef LINKER_ERROR_EXAMPLE_CONFIG_HPP
+#define LINKER_ERROR_EXAMPLE_CONFIG_HPP
+
+#include <boost/spirit/home/x3.hpp>
+
+namespace client{
+namespace parser{
+
+namespace x3 = boost::spirit::x3;
+
+
+typedef std::string::const_iterator iterator_type;
+typedef x3::phrase_parse_context<x3::ascii::space_type>::type context_type;
+
+}
+}
+
+#endif //LINKER_ERROR_EXAMPLE_CONFIG_HPP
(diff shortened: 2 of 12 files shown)


test/x3/linker_error_test/ast.hpp
```diff
@@ -0,0 +1,76 @@
+//
+// Created by lukas on 11.11.16.
+//
+
```
djowel: Odd comment. Please be consistent with the rest of the codebase. Who is lukas?

test/x3/linker_error_test/config.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef LINKER_ERROR_EXAMPLE_CONFIG_HPP
```
djowel: Same here. Please conform to the coding style; copyright at the top. Also the please follow the indentation style of the rest of the codebase.

test/x3/linker_error_test/example.cpp
```diff
@@ -0,0 +1,8 @@
+#include "example_def.hpp"
```
djowel: And again... coding style.

test/x3/linker_error_test/ast.hpp
cppljevans: I don't know a lukas.  I assume it was just copy&pasted from the original problem code
mentioned in Purpose: at top of main.cpp.

Will remove.

test/x3/linker_error_test/config.hpp
cppljevans: I could copy the ../rule1.cpp copyright, but really the code source is mentioned in, as I said
above, in main.cpp.  I guess I'll put that guys name in the copyright.

test/x3/linker_error_test/example.cpp
cppljevans: Do you mean the copyright again?  Surely the actual #include is OK.

test/x3/linker_error_test/config.hpp
djowel: Boost requires it. Actually, there's a check on commit on such things, IIRC.

test/x3/linker_error_test/ast.hpp
djowel: How about simply add these tests in one of the existing tests. Perhaps rule_tests?

cppljevans: It's only 1 test and it seemed easier to just use  a glob to select all the .cpp's for that test.
Separate .cpp's were used because that was the original author's design and it conforms
to what I thought was the accepted practice for using BOOST_SPIRIT_INSTANTIATE.
I guess I could just combine all the .cpp's into one file.  Is that what you mean?

test/x3/linker_error_test/config.hpp
cppljevans: I'll put a copyright with my name in *all* the linker_error_test files.  
Since lukas didn't copyright it, I guess that's OK.

Kojoley: I am not sure it is ok because you cannot just put a licence on code you did not write. Also this test is over-complicated IMO.

cppljevans: Good point.  I'll simplify and them putting my copyright on it would be OK.

cppljevans: An attempt was made to simplify the test here:

[complex_test](https://github.com/cppljevans/spirit-experiments/tree/BOOST_SPIRIT_INSTANTIATE/test/x3/linker_error_test)

starting with a prototype of such a simplification here:

[simple_test](https://github.com/cppljevans/spirit-experiments/tree/get_rhs/workbench/x3/BOOST_SPIRIT_INSTANTIATE_link_error)

Unfortunately, the link error didn't occur with the
simplified test, although trace output shown here:

[simple_out](https://github.com/cppljevans/spirit-experiments/blob/get_rhs/workbench/x3/BOOST_SPIRIT_INSTANTIATE_link_error/make_compare.out)

did demonstrate that with the patch, only 1 call to
`rule_definition<...>::parse` occured, but without the patch
(IOW using the existing spirit code), 2 calls to that parse
function occurred, suggesting (very strongly) that with the
existing spirit code, there is a redundant call to
`rule_definition<...>::parse`, and this is probably causing
the linke error *in some, as of yet, not understood,
circumstances*.

So, I see 2 choices:

* take the time to understand why there's no link
error with the simplified test and, with that understanding,
design a newer "a little less" simplified test that
reproduces the link error which was the initial reason for
this patch.

* just leave the simplified test code as is and replace the
  existing somewhat complicated code, which reproduces the
  link error, with the simplified test, which doesn't.

So, what should I do?

djowel: I'm with choice 1: we need to get to the bottom of this "as of yet, not understood,
circumstances." :-)

Kojoley:
Superseded by #456

PR #351, opened 2018-01-04
Prevent parse_nan() from dereferencing out of range iterator

### LINKED ISSUE #350: parse_nan() dereferences out of range iterator
In parse_nan(), after detecting the string "nan", while looking for parens, there is no check to make sure we didn't already reach the end of the range before dereferencing the iterator.

https://github.com/boostorg/spirit/blob/54d5ea1ea84013c66a138db57e6275a5aaed5035/include/boost/spirit/home/x3/numeric/real_policies.hpp#L112-L116

I believe the second `if` (line 115) should  look like:

`if (first != last && *first == '(')`

This causes asserts to trigger with debug iterators in MSVC when using std::string or std::string_view. Here's an example with a custom iterator that throws when dereferencing the end.

http://coliru.stacked-crooked.com/a/4301f38c79b25ed6

oracleoftroy:
Closes #350

### DIFF
--- a/include/boost/spirit/home/x3/numeric/real_policies.hpp
+++ b/include/boost/spirit/home/x3/numeric/real_policies.hpp
@@ -112,7 +112,7 @@ namespace boost { namespace spirit { namespace x3
             // nan[(...)] ?
             if (detail::string_parse("nan", "NAN", first, last, unused))
             {
-                if (*first == '(')
+                if (first != last && *first == '(')
                 {
                     // skip trailing (...) part
                     Iterator i = first;


djowel:
@Kojoley, what do we do when "Some checks were not successful"? Surely this PR did not case the failure, but what do I know. Hmmm...

Kojoley:
I have restarted failed jobs, they were because of some network problems at travis end plus a warning from boost.tuple you had merged patch for.

djowel:
> I have restarted failed jobs, they were because of some network problems at travis end plus a warning from boost.tuple you had merged patch for.

Oh... are we set to fail on warnings?

Kojoley:
Yes, it is good thing to request patches that does not introduce new warnings. I wish other boost libraries do the same thing.

djowel:
> Yes, it is good thing to request patches that does not introduce new warnings. I wish other boost libraries do the same thing.

👍

Kojoley:
Thanks @oracleoftroy!

oracleoftroy:
Thank you @Kojoley! It was my pleasure, and it was fun seeing all the conversations that popped up from my humble bug report and pull request.

And 👍 to efforts to avoiding warnings. I'm sure it is nearly impossible with a project with the scope of boost, but libraries that compiles with warnings at reasonable warning levels is concerning.

PR #355, opened 2018-01-24
X3: Use traits to test if container is empty

timangus:
::empty() is called on the Attribute in parse_into_container_impl. If Attribute has no ::empty(), if fails to compile. Probably instead it should use traits::is_empty(...).

### DIFF
--- a/include/boost/spirit/home/x3/core/detail/parse_into_container.hpp
+++ b/include/boost/spirit/home/x3/core/detail/parse_into_container.hpp
@@ -257,7 +257,7 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
           , Iterator& first, Iterator const& last
           , Context const& context, RContext& rcontext, Attribute& attr, mpl::true_)
         {
-            if (attr.empty())
+            if (traits::is_empty(attr))
                 return parser.parse(first, last, context, rcontext, attr);
             Attribute rest;
             bool r = parser.parse(first, last, context, rcontext, rest);


timangus:
Ah right, I'll admit I was testing using 1.66 rather than the develop branch; it looks like ee4943d would resolve my problem albeit in a different way. FWIW, my custom container is QString, and my traits are:

```cpp
namespace boost { namespace spirit { namespace x3 { namespace traits {

template<>
struct push_back_container<QString>
{
    template<typename T>
    static bool call(QString& c, T&& val)
    {
        c.push_back(std::move(val));
        return true;
    }
};

template<>
struct append_container<QString>
{
    template<typename Iterator>
    static bool call(QString& c, Iterator first, Iterator last)
    {
        c.append(first, std::distance(first, last));
        return true;
    }
};

template<>
struct is_empty_container<QString>
{
    static bool call(QString const& c)
    {
        return c.isEmpty();
    }
};

}}}}
```

timangus:
Hmm well, quickly patching ee4943d into my code doesn't fix the problem. It does seem like traits in the answer to be honest, it seems to have been in issue #287, whose fix is in the same function.

Kojoley:
> It does seem like traits is the answer to be honest

Yeah. I have asked the first question just to be sure you are using a custom container and we are not fixing some hidden problem with the patch.

timangus:
Ah right, cool. Thanks.

djowel:
👍

Kojoley:
Thanks for patch, and for tremulous!

timangus:
Haha, you're welcome! Believe it or not a small number of people still play it.

Kojoley:
Nice to hear this! I had not played it since 2011 because of server population drop (it was hard to gather together even 2x2).

PR #361, opened 2018-02-18
Added missing copy assignment operator

Kojoley:
The PR #36 does not fixed problem for C++03 and wide range of MSVC compilers.

Closes [trac 8268](https://svn.boost.org/trac10/ticket/8268).

### DIFF
--- a/include/boost/spirit/home/support/detail/hold_any.hpp
+++ b/include/boost/spirit/home/support/detail/hold_any.hpp
@@ -327,6 +327,11 @@ namespace boost { namespace spirit
             return assign(x);
         }
 #endif
+        // copy assignment operator
+        basic_hold_any& operator=(basic_hold_any const& x)
+        {
+            return assign(x);
+        }
 
         // utility functions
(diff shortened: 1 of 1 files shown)


hkaiser:
Why was this necessary? The existing `operator=()` (for both, `BOOST_HAS_RVALUE_REFS` being defined and not) were perfectly fine, iiuc.

Kojoley:
I failed to understand from the C++ standard when used defined templated assignment operator suppresses compiler generated copy constructor. From what I have tried:
  - C++11: `T&&` takes precedence, but not for older MSVC (I did not tried GCC and Clang may be they have the problem too on older versions).
  - C++03: `T const&` does not in every compiler.

The simplest workaround for all the cases is just add copy assignment operator.

hkaiser:
@Kojoley ok, fair enough. In this case I'd suggest to add an explicit `operator=(any&&)` as well (if `BOOST_HAS_RVALUE_REFS` is defined).

PR #365, opened 2018-02-20
Added fusion vector limit precheck

Kojoley:
Addresses https://svn.boost.org/trac10/ticket/10752


djowel:
makes perfect sense

Kojoley:
Actually I think this should be done on fusion side boostorg/fusion#168

Kojoley:
Closing this because it was implemented on fusion side.

PR #246, opened 2017-10-03
Karma: Perform negation in unsigned type to prevent UB

ecatmur:
When compiling with -fsanitize=undefined, `absolute_value` fails on minimal signed integer values:

```
../../../boost/spirit/home/karma/numeric/detail/numeric_utils.hpp:97:5: runtime error: negation of -9223372036854775808 cannot be represented in type 'long long int'; cast to an unsigned type to negate this value to itself
    #0 0x404e9d in boost::spirit::traits::absolute_value<long long, void>::call(long long) ../../../boost/spirit/home/karma/numeric/detail/numeric_utils.hpp:97
    #1 0x4343cd in boost::spirit::traits::absolute_value<long long, void>::type boost::spirit::traits::get_absolute_value<long long>(long long) ../../../boost/spirit/home/karma/numeric/detail/numeric_utils.hpp:149
    #2 0x42892b in bool boost::spirit::karma::any_int_generator<long long, boost::spirit::unused_type, boost::spirit::unused_type, 10u, false>::insert_int<boost::spirit::karma::detail::output_iterator<std::back_insert_iterator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, mpl_::int_<0>, boost::spirit::unused_type>, long long>(boost::spirit::karma::detail::output_iterator<std::back_insert_iterator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, mpl_::int_<0>, boost::spirit::unused_type>&, long long const&) ../../../boost/spirit/home/karma/numeric/int.hpp:212
[...]

PR #369, opened 2018-02-23
Fix integer overflow when negating smallest signed value

jwakely:
`-INT_MIN` cannot be represented as `int` so results in an undefined overflow.

Testcase from https://bugzilla.redhat.com/show_bug.cgi?id=1545092#c4

````
#include <cassert>
#include <boost/spirit/include/karma.hpp>

bool karma_to_string(std::string & str, int value) __attribute__((noinline));

bool karma_to_string(std::string & str, int value)
{
  namespace karma = boost::spirit::karma;
  std::back_insert_iterator<std::string> sink(str);
  return karma::generate(sink, value);
}

int main(int argc, char **argv)
{
  std::string out;
  karma_to_string(out, int(-2147483648));
  assert(out == "-2147483648");
  return 0;
}
````

With UBSan this shows:

/usr/include/boost/spirit/home/karma/numeric/detail/numeric_utils.hpp:90:5: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself

### DIFF
--- a/include/boost/spirit/home/karma/numeric/detail/numeric_utils.hpp
+++ b/include/boost/spirit/home/karma/numeric/detail/numeric_utils.hpp
@@ -68,7 +68,11 @@ namespace boost { namespace spirit { namespace traits
             typedef unsignedtype type;                                        \
             static type call(signedtype n)                                    \
             {                                                                 \
-                return static_cast<unsignedtype>((n >= 0) ? n : -n);          \
+                if (n >= 0)                                                   \
+                    return n;                                                 \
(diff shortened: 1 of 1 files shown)


Kojoley:
Now I understand what was PR #246 dealing with.

jwakely:
Oh I missed that one, sorry. @ecatmur's patch probably produces better code.

Kojoley:
Nothing to be sorry about, thanks for the PR!
I will merge #246 after a CI cycle.

djowel:
H @jwakely! Thanks for this patch!

PR #372, opened 2018-03-01
DOC: SGI.com Tech Archive Resources now retired [skip ci]

Kojoley:
Plus, X3 docs were missing `__stl__` def.


djowel:
👍 good catch!

PR #373, opened 2018-03-06
X3: skip_over: Remove redundant check

Kojoley:
End iterator check is redundant here because it is done in skipper.

### DIFF
--- a/include/boost/spirit/home/x3/core/skip_over.hpp
+++ b/include/boost/spirit/home/x3/core/skip_over.hpp
@@ -62,7 +62,7 @@ namespace boost { namespace spirit { namespace x3
         inline void skip_over(
             Iterator& first, Iterator const& last, Skipper const& skipper)
         {
-            while (first != last && skipper.parse(first, last, unused, unused, unused))
+            while (skipper.parse(first, last, unused, unused, unused))
                 /***/;
         }
 


djowel:
I think we should at least have a precondition assert there to be sure.

Kojoley:
We can't place an assertion here, it will fire when postskip is enabled and there is nothing to postskip.

djowel:
> We can't place an assertion here, it will fire when postskip is enabled and there is nothing to postskip.

We can add the check in the postskip so it won't fire.

Kojoley:
I don't know how it can be done without touching everything else, and it will still fire if `eps` or `attr` is the last parser in a sequence.

djowel:
> I don't know how it can be done without touching everything else, and it will still fire if eps or attr is the last parser in a sequence.

Yeah, thinking about it some more, I'm coming to the same conclusion.

djowel:
OK, you are right. It should not be done here.

cylonbrain:
Attention this change is introducing a regression in boost 1.68.0 and 1.67.0 if you use something like (as shown in some example code):
    phrase_parse(begin, end, rule, !byte);
The program than hangs in an infinite loop. I was able to fix that by using 
    phrase_parse(begin, end, rule, !eps);

PR #375, opened 2018-03-13
Karma: Fixed use after scope bug if ADT getter returns by value

Kojoley:
The `extract_from_attribute` returns a reference to a temporary
value produced by calling Fusion ADT adapted sequence getter that
returns by value.

Was reported 6 years ago https://svn.boost.org/trac10/ticket/6126


Kojoley:
I had to split default `extract_from_attribute` specialization because `fusion::result_of::at_c<Attribute, 0>` instantiation will fail if attribute is not a fusion sequence.

PR #377, opened 2018-03-15
X3: sequence: Fixed reference to temporary bug

Kojoley:


djowel:
It should be returning the reference to the fusion sequence element. Why is it not doing that?

Kojoley:
And it will if `deref` will return reference. But it can also return a value (for example `adt_attribute_proxy`), then there will be use after scope.

djowel:
So we should deal with both cases.

Kojoley:
And it already. Note that `fusion::result_of::deref` is exact return type of `deref` call.

djowel:
Oh right!

djowel:
I see. That add_reference is bogus indeed.

Kojoley:
> That add_reference is bogus

Yup :-)

PR #379, opened 2018-03-18
X3: variant: More fixes for compilation with VS2015 Update 3

RandomInEqualities:
This applies the fix in 6d6f40c3e5 to all BOOST_NOEXCEPT_IF statements in variant.hpp. I have trouble compiling a toy project with VS2015, as these noexcept if statements fails to compile.

I saw the support for VS2015 in 6d6f40c3e5, and got really execited so I checked out the develop branch to test it, unfortunately it fails for me to compile a small project that I have here:

https://gitlab.com/RandomInEqualities/Calculator

It compiles with these changes. Is the changes ok to do?

### DIFF
--- a/include/boost/spirit/home/x3/support/ast/variant.hpp
+++ b/include/boost/spirit/home/x3/support/ast/variant.hpp
@@ -47,7 +47,7 @@ namespace boost { namespace spirit { namespace x3
             boost::checked_delete(p_);
         }
 
-        forward_ast& operator=(forward_ast const& rhs) BOOST_NOEXCEPT_IF(std::is_nothrow_copy_assignable<T>{})
+        forward_ast& operator=(forward_ast const& rhs) BOOST_NOEXCEPT_IF(std::is_nothrow_copy_assignable<T>::value)
         {
             assign(rhs.get());
             return *this;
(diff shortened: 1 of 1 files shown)


Kojoley:
Hm, when I was fixing it I had come to conclusion that VS2015 has problem with rvalue references in template parameters of constexpr expressions in `noexcept` .

> I this change ok?

Yes, but it need tests since our tests suite did not catch the problem. It looks like `forward_ast` needs a fix too.

RandomInEqualities:
Year, I'll try to create a minimal example.

RandomInEqualities:
This doesn't compile:

`namespace x3 = boost::spirit::x3;`
`using A = x3::variant<x3::forward_ast<struct B>>;`
`struct B : x3::variant<A> {};`

It fails with this error: https://gist.github.com/RandomInEqualities/d45cc483cf359f3f28e9dec797b45f0a

Thinking about it i'm wondering if this even makes sense? It's trying to inherit from a variant to itself.. In the small project it's a bit more indirect that this is happening. But forward_ast<struct B> is just a pointer to B, so I guess it should work even though it might not be the best code..

Would it make sense to add a test with the above? Or insert it into the existing x3_variant.cpp test?

Kojoley:
> This doesn't compile:
>
> `namespace x3 = boost::spirit::x3;`
> `using A = x3::variant<x3::forward_ast<struct B>>;`
> `struct B : x3::variant<A> {};`

I do not understand what are your trying to do here. Why you would want a variant of a single type? Nested variants are also a very strange thing,

RandomInEqualities:
I don't want a single type, it is the minimal example that does not compile.

This is a more real minimal example:
https://gist.github.com/RandomInEqualities/d0536f4f3ebaa383f29a5d5abed78f31

In that the UnaryExpression that derives from x3::variant, is a way to parse either a double or
somthing with a +/- in front of a name. I'm not so familiar with boost spirit, so maybe that 
is a bad way to do it.

Kojoley:
Superseded by #380

PR #392, opened 2018-07-27
Drop own fp routines in favor of `boost::math`

Kojoley:
Remove own fp routines as Boost 1.40 released ~9 years ago.
The `boost::math::changesign` usage fixes trac 13531.


Kojoley:
Actually I do not know what is the purpose of using `math::changesign`. Simple `-n` makes this work perfectly (it correctly changes sign of NaN and Inf values) and compiles into a faster machine code (especially on older compilers and any msvc) https://godbolt.org/g/aVC152

djowel:
👍 I am for this PR. And yes, -n should be good, even for generic code.

hkaiser:
> Actually I do not know what is the purpose of using `math::changesign`.

IIRC, this was an optimization that was much faster for `double`/`float` than `-n`.

PR #397, opened 2018-09-15
Use endian macros from Predef

Kojoley:
The `boost/detail/endian.hpp` header is deprecated.


djowel:
It's about time. Thanks for doing this.

Kojoley:
Appveyour fail is #400 and is not related to the PR.

PR #412, opened 2018-10-26
Qi: Improve documentation for semantic actions in %=

### LINKED ISSUE #411: invalid static_cast when using _val = style assignment in semantic action Boost 1.65.1
I apologize if this comes across as a usage question rather than a bug report. I do not understand why this code does not compile and wonder whether it should:

```
qi::rule<It, std::string()> rule = qi::int_ 
   [ _val = px::bind([](int i) -> std::string { return std::to_string(2*i); }, _1) ];
```
[Full example here](http://coliru.stacked-crooked.com/a/8582cb0d28d99bdc)

Should it compile?  The example is adopted from [here](https://stackoverflow.com/questions/49859896/using-lambdas-or-external-functions-as-spirit-qi-semantic-actions) and I'll note an analogous example using [phx::new_ appears to compile](https://github.com/bingmann/2018-cpp-spirit-parsing/blob/master/spirit6_ast.cpp#L142).

If it shouldn't compile, would it be a reasonable enhancement request to support this? Thank you.

godmar:
Closes #411

### DIFF
--- a/doc/qi/nonterminal.qbk
+++ b/doc/qi/nonterminal.qbk
@@ -123,10 +123,11 @@ not defined in __qi_nonterminal__.
     [[`r = p;`]                                 [Rule definition. This is equivalent to `r %= p`
                                                 (see below) if there are no semantic actions attached
                                                 anywhere in `p`.]]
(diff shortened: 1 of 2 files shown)


Kojoley:
Thanks!

PR #413, opened 2018-10-28
to_utf8: Fixed wchar_t handling on Windows

### LINKED ISSUE #395: to_utf8 truncates characters instead of performing the conversion
https://github.com/boostorg/spirit/blob/925f40a2d0af5beed5c9a6d5882271d5d326e949/include/boost/spirit/home/support/utf8.hpp#L66

Is it permissible - convert wchar_t to char (in the case of type of the input parameter is std::wstring)?

Kojoley:
Spirit were assuming that wchar_t is 32-bit and the content is UCS-4.
It is wrong, the actual representation is implementation defined [lex.ccon]/6.
However, on most Unix platforms this assumption is valid and gives the
expected outcome, but on Windows wchar_t is 16-bit and the content is UTF-16.

Fixes #395.

### DIFF
--- a/include/boost/spirit/home/support/utf8.hpp
+++ b/include/boost/spirit/home/support/utf8.hpp
@@ -67,6 +67,44 @@ namespace boost { namespace spirit
         }
         return result;
     }
+
+    // Assume wchar_t content is UTF-16 on Windows and UCS-4 on Unix
+#if defined(_WIN32) || defined(__CYGWIN__)
+    inline utf8_string to_utf8(wchar_t value)
+    {
+        utf8_string result;
+        typedef std::back_insert_iterator<utf8_string> insert_iter;
+        insert_iter out_iter(result);
+        utf8_output_iterator<insert_iter> utf8_iter(out_iter);
+
+        u16_to_u32_iterator<wchar_t const*, ucs4_char> ucs4_iter(&value);
+        *utf8_iter++ = *ucs4_iter;
+
+        return result;
+    }
+
+    inline utf8_string to_utf8(wchar_t const* str)
+    {
+        utf8_string result;
+        typedef std::back_insert_iterator<utf8_string> insert_iter;
+        insert_iter out_iter(result);
+        utf8_output_iterator<insert_iter> utf8_iter(out_iter);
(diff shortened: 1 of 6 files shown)


djowel:
wchar_t is implementation defined. If the code somehow assumes that, then it is a bug.

Kojoley:
What are alternatives? Remove `to_utf8` entirely? It is been here for 10 years and assumes that wchar_t is 32-bit UCS-4. I would be glad if you can fix this somehow without assuming any implementation details of wchar_t. Note that `std::codecvt` is deprecated in C++17 http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0618r0.html.

djowel:
How about we support only a certain subset of types: those that we know work and is generally useful for typical platforms such as windows and unix? Assert the size required perhaps.

Kojoley:
> How about we support only a certain subset of types: those that we know work and is generally useful for typical platforms such as windows and unix?

The problem is we cannot know what can be passed here, especially in C++98 because there is no char16_t and char32_t types.

> Assert the size required perhaps.

This function is only used in `parser.what()` (I have posted about this in https://github.com/boostorg/spirit/issues/395#issuecomment-433712373, please read it), no one noticed before that it does something strange with wchar_t on Windows because it is does not affect parsing in any way and its result is only seen in parser debugging with `BOOST_SPIRIT_QI_DEBUG`. I think it is not a good way to throw asserts from things that are cosmetic and a size assert will not help, possibly constexpr function could check what the actual wchar_t representation is, but again I do not think it is needed. Also I have added a test and when it will fail somewhere then things might be reconsidered.

djowel:
I see. Well, I think we should simplify and just require exact representations to begin with. All platforms has some conversion support for it anyway. I think your last paragraph in your comment is the way to go.

Kojoley:
I am sorry I do not get you. If would be glad if you fix this if you know how it should be done the right way.

djowel:
Sorry for the misunderstanding. To be clear, I think your solution here is OK.

Kojoley:
I merge it then?

PR #427, opened 2018-12-17
Qi: rule: Added asserts for mismatching skipper types

### LINKED ISSUE #426: Comment skipper not working in current Boost Spirit
I am getting these kinds of errors following some of the online examples, on version ``Boost v1.68.0``:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2664	'bool boost::function4<R,T0,T1,T2,T3>::operator ()(T0,T1,T2,T3) const': cannot convert argument 4 from 'const Skipper' to 'T3'	Kingdom.Protobuf.Agents	d:\dev\boost.org\boost_1_68_0\boost\spirit\home\qi\nonterminal\rule.hpp	304	
```

Such a skipper is not a complex issue, but something is not right, I think:

```C++
template<typename It>
struct ProtoCommentSkipper : qi::grammar<It> {

    ProtoCommentSkipper() : ProtoCommentSkipper::base_type(skip) {

        using qi::char_;
        using qi::eol;
        using qi::ascii::space;

        skip %= space
            | ("//" >> *(char_ - eol) >> eol)
            | ("/*" >> *(char_ - "*/") >> "*/")
            ;
    }

private:
    qi::rule<It> skip;
};
```

And in usage:

```C++
template<typename It, typename Skipper = qi::space_type>
struct ProtoGrammar : qi::grammar<It, ast::proto_t(), Skipper> {
    // ...
    using skipper_type = Skipper;

    ProtoGrammar() : ProtoGrammar::base_type(start) {
        // ...
    }

private:
    // ...
    qi::rule<It, ast::proto_t(), skipper_type> start;
};
```

The only difference is that I sub in the ``ProtoCommentSkipper`` and I get the errors. Instead of:

```C++
using iterator = std::string::const_iterator;
using proto_grammar = ProtoGrammar<std::string::const_iterator>;
// Where s is std::string ...
iterator begin = s.begin(), end = s.end();
const auto parse_tried = qi::parse(begin, end, proto_grammar{}, x);
```

I am trying this:

```C++
using iterator = std::string::const_iterator;
using proto_comment_skipper = ProtoCommentSkipper<std::string::const_iterator>;
using proto_grammar = ProtoGrammar<std::string::const_iterator, ProtoCommentSkipper<std::string::const_iterator>>;
iterator begin = s.begin(), end = s.end();
const auto parse_tried = qi::parse(begin, end, proto_grammar{}, proto_comment_skipper{}, x);
```

Seems like a possible duplicate of [Boost library spirit failed to build](/boostorg/spirit/issues/399), however, do you happen to have a simple example of a comment parser? Using a *Spirit Qi Confix*, for instance?

Kojoley:
It looks like mismatching skipper types and `parse`/`phrase_parse` calls are
quite popular mistakes, and the actual problem is not clear for users despite
that there is a comment near the call that will fail and the hint should be
found by following to the file and line that compiler reports the error on.

Closes #426.

### DIFF
--- a/include/boost/spirit/home/qi/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/qi/nonterminal/rule.hpp
@@ -12,10 +12,12 @@
 #endif
 
 #include <boost/assert.hpp>
+#include <boost/static_assert.hpp>
 #include <boost/config.hpp>
 #include <boost/function.hpp>
 #include <boost/mpl/vector.hpp>
 #include <boost/type_traits/add_reference.hpp>
+#include <boost/type_traits/is_convertible.hpp>
 #include <boost/type_traits/is_same.hpp>
 
 #include <boost/fusion/include/vector.hpp>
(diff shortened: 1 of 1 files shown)


mwpowellhtx:
I'm not sure what you mean:

> ... mismatching skipper types ...

That's not what's going on at all. I tried furnishing a skipper instance during my troubleshooting of the issue, yes; but that is by no means the root cause.

PR #435, opened 2019-01-08
X3: fix unused parameter warnings in position_tagged

Xeverous:


Kojoley:
Thanks. However this is not the only unused parameter warning.

Xeverous:
@Kojoley got any list of them? I fixed the stuff that just happened to trigger on my own projects

Also, do you care about fallthrough warnings? I don't like any warnings popping out not of my code when building but I'm not sure how to approach them. Do you have any convention how to remove "may fallthrough" warnings?

Kojoley:
> @Kojoley got any list of them?

I have a patch in my backlog but did not merged it yet.

> Do you have any convention how to remove "may fallthrough" warnings?

There is `BOOST_FALLTHROUGH` macro for this purpose. In C++17+ code you can use `[[fallthrough]]`.

The problem with fixing warnings is that you need to fix bunch of libraries and there is no point in fixing it in Spirit until they are not fixed in underlying libraries. Some time ago I did a bunch of such fixes, for some libraries it took more than a year to get the fix merged.

Kojoley:
I even remember that warning, though I might not filed a PR :) Good luck!

PR #437, opened 2019-01-13
X3: Fixed wrong parse_rule signature instantiation

Kojoley:
When a rule has no attribute (leaves `Attribute` template parameter as default)
the `BOOST_SPIRIT_INSTANTIATE` macro instantiates a `parse_rule` helper function
with a wrong signature because `parse`, `phrase_parse`, and `skip_over` pass as
the attribute `unused` variable which deduces to `unused_type const&` type, but
`BOOST_SPIRIT_INSTANTIATE` instantiates `parse_rule` with `unused_type&` type.

Used a trick to change `parse_rule` signature to take unsued_type by value when rule has no attribute (the signature `parse_rule(..., unsued_type)`).

### DIFF
--- a/include/boost/spirit/home/x3/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/rule.hpp
@@ -13,6 +13,7 @@
 #include <boost/preprocessor/variadic/to_seq.hpp>
 #include <boost/preprocessor/variadic/elem.hpp>
 #include <boost/preprocessor/seq/for_each.hpp>
+#include <type_traits>
 
 #if !defined(BOOST_SPIRIT_X3_NO_RTTI)
 #include <typeinfo>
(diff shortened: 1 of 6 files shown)


cppljevans:
On 1/13/19 10:08 AM, Nikita Kniazev wrote:
> When a rule has no attribute the `BOOST_SPIRIT_INSTANTIATE` macro instantiates
> a `parse_rule` function with a wrong signature because `parse`, phrase_parse`,
> and `skip_over` use `unused` variable which has `unused_type const` type, while
> default template parameter for `rule` is non-const.
> 
> There are three possible solutions:
> 
>    1. Change default template parameter type of `rule` to `unused_type` const.
>    2. Change `unused` variable type to non-const.
>    3. Do not use `unused` for attributes, instead create a local variable
>       of `unused_type` and pass it as the attribute.
> 
> Forbidding non-const attributes is not a solution as it will require a check in
> every parser and is simple to miss at adding new parsers.

So, which solution was chosen?  From looking at detail/rule.hpp:29 here:

https://github.com/boostorg/spirit/pull/437/files#diff-632a9a805644abb7caf7f726e70249a1R29

I'm guessing 1 was chosen.  What's the advantage of 1 over 2?

[snip]

-regards,
Larry

Kojoley:
> I'm guessing 1 was chosen.

The PR did not yet merged, but yes, currently it implements the first way.

>  What's the advantage of 1 over 2?

It will not change the meaning of `unused` for everybody and will allow later to make it `constexpr`.

cppljevans:
On 1/13/19 12:15 PM, Nikita Kniazev wrote:
>> I'm guessing 1 was chosen.
> 
> The PR did not yet merged, but yes, currently it implements the first way.
> 
[snip]
Wouldn't this change the value of has_attribute here:

https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/x3/nonterminal/rule.hpp#L46

How are you sure this change is harmless?

-regards,
Larry

Kojoley:
Updated the PR with other solution. Used a trick to change `parse_rule` signature to take `unsued_type` by value when rule has no attribute (the signature `parse_rule(..., unsued_type)`).

I think that `parse_rule` should NOT deduce attribute type (the attribute transformation must occur before it, not after), but this is out of this PR scope.

djowel:
LGTM!

Kojoley:
This is a bullet proof solution, it does not care about passed variable constness. I do not find the code complicated, it is just a bit ugly as any workaround. There are a lot of other places except the ones mentioned above where `unused` is passed as attribute.

Issue #434, opened 2019-01-08
X3: Sequence in variant parser with container attribute


Xeverous:
This problem originates from https://stackoverflow.com/questions/48390872 (not my question).

SO attempt: http://coliru.stacked-crooked.com/a/56007d3f997b9748

My attempt: http://coliru.stacked-crooked.com/a/3365b20994ce460e

It seems that `\` is basically impossible to obtain, it's always skipped.

Also, current X3 doc states that the attribute of `lit(c)` is unused and the attribute of `char_(c)` is the type of `c`. This does not seem to be the case as the first code sample has no difference between `lit(c)` and `char_(c)` versions.

Kojoley:
MWE:
```cpp
#include <boost/spirit/home/x3.hpp>
#include <iostream>
#include <string>

namespace x3 = boost::spirit::x3;

int main()
{
    char const* s = "aaabcd", *const end = s + std::strlen(s);
    std::string x;
    if (!parse(s, end, *((x3::char_('a') >> x3::char_) | x3::char_), x))
        std::cout << "failed\n";
    else
        std::cout << "result=" << x << '\n';
}
```
Output:
```
result=abcd
```
https://wandbox.org/permlink/SSv343kksXESwQyU

The fun part is that `&` operator forces it to synthesize the value (while it should not)

```cpp
#include <boost/spirit/home/x3.hpp>
#include <iostream>
#include <string>

namespace x3 = boost::spirit::x3;

int main()
{
    char const* s = "aaabcd", *const end = s + std::strlen(s);
    std::string x;
    if (!parse(s, end, *((&x3::char_('a') >> x3::char_) | x3::char_), x))
        std::cout << "failed\n";
    else
        std::cout << "result=" << x << '\n';
}
```
Output:
```
result=aaabcd
```

@djowel what is the difference between `&a` and `omit[a]`? Ain't they the same thing?

Kojoley:
The actual problem should is sequence in variant parser with container attribute.
The `((x3::char_ >> x3::char_) | x3::char_)` parser does not even compile with container attributes.
With repeated parsers the sequence just returns the last value:
```cpp
#include <boost/spirit/home/x3.hpp>
#include <iostream>
#include <string>

namespace x3 = boost::spirit::x3;

int main()
{
    char const* s = "aaabcde", *const end = s + std::strlen(s);
    std::string x;
    if (!parse(s, end, *((x3::char_ >> x3::char_ >> x3::char_) | x3::char_), x))
[...]

PR #439, opened 2019-01-15
X3:  Removed sequence into plain parsing

### LINKED ISSUE #438: X3: Sequence will parse into any attribute
```cpp

int main()
{
    char const* s = "";
    
    int i;
    parse(s, s, x3::int_ >> x3::int_, i);

    boost::optional<int> o;
    parse(s, s, x3::int_ >> x3::int_, o);

    boost::variant<int, long> v;
    parse(s, s, x3::int_ >> x3::int_, v);
}
```

I was expecting this to not compile, but it does.

The code that responsible for this: https://github.com/boostorg/spirit/blob/a36077efbd3b91bc7dbfad211bcf4533aa137159/include/boost/spirit/home/x3/operator/detail/sequence.hpp#L328-L390

Basically it is wrong and/or redundant. I do not know what was original intentions, but what actually it does - pass the attribute to all underlying parsers of a sequence. I tried to remove it and did not find a test case that will fail.

Kojoley:
Closes #438.


djowel:
I assume all tests pass? Just be sure this does not affect rules. I think you are in the right direction, but I'm somewhat worried.

Kojoley:
> I assume all tests pass?  I think you are in the right direction, but I'm somewhat worried.

Yes, and examples compile too. We also have CI set up for this purpose.

> Just be sure this does not affect rules.

The commit where the code come 235359a45c937daf2c96ce96e869d7e57604eb01 seems to be about tuple elements pass-through, and rules itself is just a parser for sequences.

> I think you are in the right direction, but I'm somewhat worried.

I expect this change to break the code that is already silently broken (like #434).


@cppljevans I prefer not to make unrelated changes. I agree that my naming may be confusing, you can open a PR for this, but the `pass` means something like pass-through or forwarding.

cppljevans:
> I would like to add a test case that ensures `eps >> eps` does NOT compile with `tuple<>` attribute, but I do not want to make the current code more complex just to disable it and compile fail tests are too slow (end requires a separate file for every case).

If it does NOT compile, then I'm puzzled because I'm getting it to compile.  With:

````cpp
#include <boost/fusion/container/vector.hpp>
int main()
{
    {
        using boost::spirit::x3::eps;
    //Copy&past from:
    //  https://github.com/boostorg/spirit/pull/439/commits/2235da2416b1ffbaa167aa2e9ebc5aeb98eab11b
        char const* const s = "";
        boost::fusion::vector<> a;
        parse(s, s, eps >> eps, a);
    }
    return 0;
}
````
It compiles with your patch:

````
/usr/bin/clang++ -c -O0 -gdwarf-2  -std=c++17 -ftemplate-backtrace-limit=0 -fdiagnostics-show-template-tree -fno-elide-type -fmacro-backtrace-limit=0     -I../../x3-Removed_sequence_into_plain_parsing-zip/include -I/home/evansl/prog_dev/boost/releases/ro/boost_1_69_0    -ftemplate-depth=200  test-sequence-eps2.cpp -MMD -o /tmp/build/clangxx6_0_pkg/boost/releases/ro/boost_1_69_0/sandbox/kojoley/test/x3-Removed_sequence_into_plain_parsing/test-sequence-eps2.o 
/usr/bin/clang++    /tmp/build/clangxx6_0_pkg/boost/releases/ro/boost_1_69_0/sandbox/kojoley/test/x3-Removed_sequence_into_plain_parsing/test-sequence-eps2.o   -o /tmp/build/clangxx6_0_pkg/boost/releases/ro/boost_1_69_0/sandbox/kojoley/test/x3-Removed_sequence_into_plain_parsing/test-sequence-eps2.exe
/tmp/build/clangxx6_0_pkg/boost/releases/ro/boost_1_69_0/sandbox/kojoley/test/x3-Removed_sequence_into_plain_parsing/test-sequence-eps2.exe 

Compilation finished at Thu Jan 17 07:41:17
````
Your patch is in the:

  ../../x3-Removed_sequence_into_plain_parsing-zip/include

directory.

The -zip directory was downloaded early yesterday from:

https://github.com/boostorg/spirit/tree/a6132ae361b5f4d3480d23200ea8d10b8911ff55

What might I be doing wrong?

TIA.

-regards,
Larry

Kojoley:
> What might I be doing wrong?

Possibly I was not clear enough, I try to put emphasis on my previous message

> **I would like** to add a test case that ensures `eps >> eps` does NOT compile with `tuple<>` attribute, **but I do not want to make the current code more complex just to disable it** ...

I consider it as a bug and it is out of this PR scope. I have created #440 about it.

djowel:
Wonderful!

PR #441, opened 2019-01-20
Tests: Added precompiled headers

Kojoley:
Not intrusive. Significantly speeds up tests on MSVC (for about 3 times). Finally solves exceeding build time limits on CI with MSVC compilers.

Did not set up PCH for X3, it is fast as-is (tooks only about 2 minutes). Repository tests reuse PCH from Qi/Karma tests.


djowel:
So cool!

PR #443, opened 2019-01-23
Replace mpl::find_if+is_same with find

Kojoley:
From `mpl::find` docs:
> `typedef find<s,t>::type i;`
> Semantics: Equivalent to `typedef find_if<s, is_same<_,t> >::type i;`


djowel:
duh! of course it is. good find (pun intentional)! :-)

PR #446, opened 2019-02-01
X3: fix for attribute copying bug

### LINKED ISSUE #444: Rules do not pass-through attributes
IMO this is an unwanted bug, as copies of various AST subparts can be expensive. It's also nowhere stated in the documentation that types need to be copyable.

I don't have an MCVE now (ask me if you want any, I will try to reduce my project as long as it reproduces). The error message (template backtrace) points to spirit implementation:

    ../src/parser/ast.hpp:243:8:   required from 'static bool boost::spirit::x3::detail::rule_parser<Attribute, ID>::call_rule_definition(const RHS&, const char*, Iterator&, const Iterator&, const Context&, ActualAttribute&, ExplicitAttrPropagation) [...]
    [...]/boost/spirit/home/x3/nonterminal/rule.hpp:61:39:   required from 'bool boost::spirit::x3::rule_definition<ID, RHS, Attribute, force_attribute_>::parse(Iterator&, const Iterator&, const Context&, boost::spirit::x3::unused_type, Attribute_&) const [...]
    ../src/parser/grammar_def.hpp:256:1:   required from 'bool fs::parser::parse_rule(fs::parser::grammar_type, Iterator&, const Iterator&, const Context&, Attribute&) [...]
    ../src/parser/grammar.cpp:13:1:   required from here
    [...]/mingw64/include/c++/8.2.0/bits/stl_construct.h:75:7: error: use of deleted function 'fs::parser::ast::action_expression::action_expression(const fs::parser::ast::action_expression&)'

digging the implementation (`x3/nonterminal/rule.hpp`, lines 56-66):

        template <typename Iterator, typename Context, typename Attribute_>
          , Context const& context, unused_type, Attribute_& attr) const
        {
            return detail::rule_parser<attribute_type, ID>
                ::call_rule_definition(
                    rhs, name, first, last
                  , context
                  , attr
                  , mpl::bool_<force_attribute>());
        }

then (`x3/nonterminal/detail/rule.hpp`, lines 295-349)

        template <typename RHS, typename Iterator, typename Context
          , typename ActualAttribute, typename ExplicitAttrPropagation>
        static bool call_rule_definition(
            RHS const& rhs
          , char const* rule_name
          , Iterator& first, Iterator const& last
          , Context const& context, ActualAttribute& attr
          , ExplicitAttrPropagation)
        {
            boost::ignore_unused(rule_name);

            typedef traits::make_attribute<Attribute, ActualAttribute> make_attribute;

            // do down-stream transformation, provides attribute for
            // rhs parser
            typedef traits::transform_attribute<
                typename make_attribute::type, Attribute, parser_id>
            transform;

            typedef typename make_attribute::value_type value_type;
            typedef typename transform::type transform_attr;
            value_type made_attr = make_attribute::call(attr); // HERE IS THE COPY
            transform_attr attr_ = transform::pre(made_attr);

            bool ok_parse
              //Creates a place to hold the result of parse_rhs
              //called inside the following scope.
              ;
            {
             // Create a scope to cause the dbg variable below (within
             // the #if...#endif) to call it's DTOR before any
             // modifications are made to the attribute, attr_ passed
             // to parse_rhs (such as might be done in
             // traits::post_transform when, for example,
             // ActualAttribute is a recursive variant).
    #if defined(BOOST_SPIRIT_X3_DEBUG)
                context_debug<Iterator, transform_attr>
                dbg(rule_name, first, last, attr_, ok_parse);
    #endif
                ok_parse = parse_rhs(rhs, first, last, context, attr_, attr_
                   , mpl::bool_
                     < (  RHS::has_action
                       && !ExplicitAttrPropagation::value
                       )
                     >()
                  );
            }
            if (ok_parse)
            {
                // do up-stream transformation, this integrates the results
                // back into the original attribute value, if appropriate
                traits::post_transform(attr, std::forward<transform_attr>(attr_));
            }
            return ok_parse;
        }

`x3/support/traits/make_attribute`

    template <typename Attribute>
    struct make_attribute_base
    {
        static Attribute call(unused_type)
        {
             // synthesize the attribute/parameter
            return Attribute();
        }

        template <typename T>
        static T& call(T& value)
        {
            return value; // just pass the one provided
        }
    };

The core of the problem lies on these 2 lines

        value_type made_attr = make_attribute::call(attr);
        transform_attr attr_ = transform::pre(made_attr);

`make_attribute::call()` default-constructs the attribute if originally it was unused, but copies it instead if it was. At first I though that the library wants to preserve the state of the attribute (in case it's already partially filled) but ... no, **the copy that is just made is then passed to `transform::pre` which doesn't do anything with the argument and returns a default-constructed object of type `transform_attr`. `made_attr` is never used anymore which makes it essentially a useless copy**. At the end, `attr` (the orginal user object) is move-constructed from `attr_`.

Xeverous:
X3 has used type transformations that unintentionally made copies
of the suplied attribute. This prevented non-copyable AST types
and had a potential performance impact.

- `value_initialize` trait: removed `boost::value_initialized` dependency (which makes sense only for pre-11 standards), the code now uses C++11 brace initialization
- `transform_attribute::pre` now uses `value_initialize` trait
- removed all usages of (now redundant) `make_attribute` trait, its
functionality is covered by `transform_attribute`
- removed `make_attribute` and associated includes

This fixes #444 

___

question related to the code changes: At the end of `rule_parser::call_rule_definition` there is `traits::post_transform(attr, std::forward<transform_attr>(attr_));`. `transform_attr` can be a reference type - is then such forward intended? Shouldn't we `std::forward<std::remove_reference_t<transform_attr>>(attr_)` ?


include/boost/spirit/home/x3/nonterminal/detail/transform_attribute.hpp
```diff
@@ -22,7 +23,10 @@ namespace boost { namespace spirit { namespace x3
     {
         typedef Transformed type;
 
-        static Transformed pre(Exposed&) { return Transformed(); }
+        static Transformed pre(Exposed&)
+        {
+            return traits::value_initialize<Transformed>::call();
```
Kojoley: Hm, I thought `make_attribute` uses `value_initialize`, maybe we do not need it at all. Revert this for now.

include/boost/spirit/home/x3/support/traits/value_traits.hpp
```diff
@@ -9,16 +9,14 @@
...
-
 namespace boost { namespace spirit { namespace x3 { namespace traits
 {
     template <typename T, typename Enable = void>
     struct value_initialize
     {
         static T call()
         {
-            return boost::value_initialized<T>();
+            return {};
```
Kojoley: Please file this as a separate PR. The change does not need addition tests, while other changes in this PR do.

Xeverous:
@Kojoley tests: how would you like tests to be done? I see 2 options:

- use the minimal working example from referenced issue, this will cause test compilation to fail in case of regression
- write the class so that the copying is possible but the class notes that it has been copied (more complex code (need to avoid copy elision) but test will pass/fail instead)

Kojoley:
Compile tests are preferred. You can use some known move only type like `unique_ptr` if it is easier for you (we do not need to proof that `unique_ptr` is move only). There are attribute synthesizing tests in `rule3.cpp`, it is ok to tune them to use move only type.

Currently the test suit fails on your PR, start by fixing them.

Kojoley:
> question related to the code changes: At the end of rule_parser::call_rule_definition 
> there is traits::post_transform(attr, std::forward<transform_attr>(attr_));. transform_attr can be a reference type - is then such forward intended? Shouldn't we std::forward<std::remove_reference_t<transform_attr>>(attr_) ?

No, it is fine as is. The forward returns lvalue if template parameter was lvalue and rvalue if template parameter was rvalue reference or value type - that's exactly what we want to pass to `post`.

include/boost/spirit/home/x3/nonterminal/detail/transform_attribute.hpp
Kojoley: Qi's `make_attribute` uses `value_initialize`, that's where my confusion came from. Currently I do not know why Qi uses it (from `value_initialized` documentation it looks like it used wrongly), and Qi does not have `value_initialize` so no need to add complexity until there is a real need for it.

Xeverous:
What is the point of these specializations?

    template <typename Attribute>
    struct transform_attribute<unused_type, Attribute>
      : transform_attribute<unused_type, unused_type> {};

    template <typename Attribute>
    struct transform_attribute<unused_type const, Attribute>
      : transform_attribute<unused_type, unused_type> {};

Both end making `unused_type` when the `Transformed` was supposed to be something that is not unused. They cause current patch to fail the build, because in the `parse()` implementation `unused_type` is returned from `_attr` which is obviously wrong.

I don't get these specializations. If I understand correctly what we want to do (assuming combined `make_attribute` and `transform_attribute` impelmentation) is:

default: `type = Transposed`

- spec when `Exposed == Transformed`: `type = Transposed&`
- spec when `Exposed == unused_type`: `type = Transposed`
- spec when `Transformed = unused_type`: `type = unused_type`

include/boost/spirit/home/x3/core/action.hpp
```diff
@@ -82,14 +81,12 @@ namespace boost { namespace spirit { namespace x3
...
-            typedef traits::make_attribute<attribute_type, unused_type> make_attribute;
             typedef traits::transform_attribute<
-                typename make_attribute::type, attribute_type, parser_id>
+                unused_type, attribute_type, parser_id>
             transform;
 
             // synthesize the attribute since one is not supplied
-            typename make_attribute::type made_attr = make_attribute::call(unused_type());
-            typename transform::type attr = transform::pre(made_attr);
+            typename transform::type attr = transform::pre(unused_type());
```
Kojoley: I think we can replace this with `attribute_type attr{};`. No one should specialize something like `transform_attribute<unused_type, int>`, it would be nonsense.

Xeverous:
ok, so we can start implementing the class from scratch - it is not used anywhere else in the library besides action and rule

assuming the task is to just provide any meaningful attribute for the parsing, the implementation can go like this:

- Exposed: the type that user provided (may be unused)
- Transformed: the desired type that is needed for parsing

default case:
- `type = Transformed`
- `pre(Exposed&)` returns value-constructed object of type `Transformed`
- `post(Exposed&, Transformed&&)` moves from `Transformed` to `Exposed`

case when `Exposed == Transformed` (optimize by using references)
- `type = Transformed&`
- `pre(Exposed&)`: return reference to the provided object
- `post()`: no-op

case when `Exposed = unused_type`
- `type = Transformed`
- `pre(unused_type)` returns value-constructed object of type `Transformed`
- `post(unused_type, const Transformed&)`: no-op

other cases (reference ones and twice unused) - I don't see the point right now

What do you think?

include/boost/spirit/home/x3/nonterminal/detail/transform_attribute.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include <boost/spirit/home/x3/support/traits/transform_attribute.hpp>
 #include <boost/spirit/home/x3/support/traits/move_to.hpp>
+#include <boost/spirit/home/x3/support/traits/value_traits.hpp>
```
Xeverous: accidental stray edit, remove this when adding tests

cppljevans: I'm getting error when compiling test/char_class.cpp:
```bash
xeverous/spirit/test/x3/char_class.cpp:110:26: warning: illegal character encoding in string literal [-Winvalid-source-encoding]
        BOOST_TEST(test("<E9>", alpha));
                         ^~~~
```
Is that expected?

Kojoley: Yes, read a header comment in `char_class.cpp`.

Kojoley:
FYI I opened PR #448 that removes redundant and wrong specializations.

Kojoley:
It seems that the things are more complicated and there was a hidden bug.

```cpp
#include <boost/core/lightweight_test.hpp>
#include <boost/fusion/include/adapt_struct.hpp>

struct Foo
{
    Foo() : i{}, j{} { ++constructed; }

    static int constructed;
    int i, j;
};

int Foo::constructed = 0;

BOOST_FUSION_ADAPT_STRUCT(Foo, i, j)


int main()
{
    {
        using boost::spirit::x3::int_;
        using boost::spirit::x3::rule;

        auto p = rule<class r, Foo>{} = '{' >> int_ >> ':' >> int_ >> '}';

        auto s = "{123:456}", e = s + std::strlen(s);
        BOOST_TEST(parse(s, e, p));
        BOOST_TEST_EQ(Foo::constructed, 0);
    }

}
```

https://wandbox.org/permlink/Iznpoky24gcTeL2H

`make_attribute` synthesizes the attribute when it is not needed, and then `transform_attribute` does not let it pass to the parser, but with your last change (and in #448) it now propagates to the parser (what will case performance degradation).

Kojoley:
I have a series of PR that will fix you problem. The first is #449.

Xeverous:
What do you think about adding or modifying existing test to make one with non-copyable AST?

PR #448, opened 2019-02-02
Removed wrong and redundant transform_attribute specializations

Kojoley:
The wrong specializations were allowing to assign a parse to a rule with
incompatible attribute types.


Kojoley:
Hm, I caught a bug in tests because rules started to synthesize attributes, I should not remove these specializations and will revert that part, but it is a nice finding.

Kojoley:
Omg, they were synthesizing, but in `make_attribute` and then `transform_attribute` just did not let it go into parser. So... we need to fix this bug first.

Xeverous:
There definitely is some issue with make/transform attribute traits. A lot of my changes seem to break some tests. I'm not sure what these traits should do exactly by default.

PR #451, opened 2019-02-03
unused_type tweaks

Kojoley:
Added basic tests and on C++11 made the type trivial and all operations with it constexpr.

### DIFF
--- a/include/boost/spirit/home/support/unused.hpp
+++ b/include/boost/spirit/home/support/unused.hpp
@@ -15,11 +15,6 @@
 #include <boost/config.hpp>
 #include <boost/mpl/bool.hpp>
 
-#if defined(BOOST_MSVC)
-# pragma warning(push)
-# pragma warning(disable: 4522) // multiple assignment operators specified warning
-#endif
-
 ///////////////////////////////////////////////////////////////////////////////
 namespace boost { namespace spirit
 {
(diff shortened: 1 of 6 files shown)


cppljevans:
Why make unused_type constexpr when just providing a constexpr unused_type::CTOR enables
even non-const unused_type to be used in a constexpr expression?

Kojoley:
Do you have any cons?

Kojoley:
What it restricts you to do? You really expect any of those functions/methods to be changed?
I will push X3 to be almost all constexpr so it will be possible to make parsers and parse text at compile time, the `unused_type` is a beginning as it used everywhere and I do not want to make it constexpr in several of iterations.

cppljevans:
On 2/3/19 4:31 PM, Nikita Kniazev wrote:
> What it restricts you to do? You really expect any of those functions/methods to be changed?
No.  Just as I wouldn't expect any functions/methods to change if
unused declaration were changed to:

   unused_type unused;

where unused_type had a constexpr CTOR.  As I said, it may be harmless 
but redundant; hence, unnecessary.

> I will push X3 to be almost all constexpr so it will be possible to make parsers and parse text at compile time, the `unused_type` is a beginning as it used everywhere and I do not want to make it constexpr in several of iterations.
> 
According to compile tests I've done with declarations similar to the 
above `unused` declaration, this is possible:

   struct tag_t { constexpr tag_t(){} };
   tag_t tag_nc; //tag_nonconst
   constexpr int result_nc=f(tag_nc);

where, of course, tag_t is used in place of unused_type.

Hence, I don't see how such a declaration would prevent you from
accomplishing your goal to "make parsers and parse text at compile time".

Please provide code examples actually supporting your contention.

-regards,
Larry

Kojoley:
I do not understand what you want to demonstrate with the example, it shows nothing.

You still did not provide cons. What do you want from me, to close the PR? No problem, I do not want to have meaningless discussions about your fears. Every change I making to Spirit is hard for me and making it way to the repository even harder as I have to judge a lot of consequences. And when I receive 'worrying' like 'harmless
but redundant; hence, unnecessary'.

PR #458, opened 2019-02-09
Remove broken 1.0#INF parser

### LINKED ISSUE #415: 1.0#... does not work and never worked but has code for it
Update: Was already reported on trac 5 years ago https://svn.boost.org/trac10/ticket/8699.

Real parsers fail on infinity values in form `1.0#INF` and it never ever worked (in both Qi and X3). The only mention that Spirit parses things like that was buried in [`RealPolicies` documentation](https://www.boost.org/doc/libs/1_68_0/libs/spirit/doc/html/spirit/qi/reference/numeric/real.html#spirit.qi.reference.numeric.real._code__phrase_role__identifier__realpolicies__phrase___code__expression_requirements).

The current code
https://github.com/boostorg/spirit/blob/cf631bf68f0efdeabdfba24a0dea515ece255298/include/boost/spirit/home/qi/numeric/detail/real_impl.hpp#L300-L315
accepts values in undocumented form like `1INF` and `1.INF`, but it fails with mantissa - i.e., `1.0INF`.

Fixing it to allow `1.0#INF` scares me because it will be backward-incompatible change for users that rely on current behavior, it will break parsers like `double_ >> "#..."`. I think this code need to be removed, as it never accepted values in documented form and has undocumented implications. Removing the code will make real parser slightly faster and will remove custom real requirement for `traits::is_equal_to_one` (partly solves the problem mentioned in #163).

```cpp
        BOOST_TEST(test("1#INF", double_));    // Fails
        BOOST_TEST(test("1.#INF", double_));   // Fails
        BOOST_TEST(test("1.0#INF", double_));  // Fails
        BOOST_TEST(test("1INF", double_));     // Ok, but must not
        BOOST_TEST(test("1.INF", double_));    // Ok, but must not
        BOOST_TEST(test("1.0INF", double_));   // Fails
```

Kojoley:
It was poorly documented and never worked correctly. None of atof, strtof, and
input streams are accepting such values. Fixing it would be a bigger breaking
change than removing it.

Fixes #415, addresses #163 and https://svn.boost.org/trac10/ticket/8699

### DIFF
--- a/doc/qi/numeric.qbk
+++ b/doc/qi/numeric.qbk
@@ -682,10 +682,6 @@ imposed by custom policies.
                                             [Negate `n` if `b` is `true`. Default implementation
                                             is provided for `float`, `double` and `long double`.]]
 
-    [[`boost::spirit::traits::is_equal_to_one(n)`]
-                                            [Return `true` if `n` is equal to `1.0`. Default implementation
-                                            is provided for `float`, `double` and `long double`.]]
-
 ]
 
 [note The additional spirit real number traits above are provided to
@@ -716,7 +712,7 @@ corresponds to the following grammar:
         ;
 
     nan
-        =   -lit("1.0#") >> no_case["nan"]
+        =   no_case["nan"]
             >> -('(' >> *(char_ - ')') >> ')')
         ;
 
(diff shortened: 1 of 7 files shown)


djowel:
Let's do it.

PR #460, opened 2019-02-12
X3: Replaced make_attribute with direct attribute construction

Kojoley:
If the attribute has a general way to be constructed it should already have a
default constructor.


djowel:
These are customization points. There's a corresponding CP in Qi, again IIRC. I don't think this should be removed.

Kojoley:
> These are customization points.

`make_attribute` is not advertised as a customization point anywhere and it has a strange signature to be one of.

> There's a corresponding CP in Qi

Yes there is the same thing in Qi and there are also questions about it (Karma is behaving weird IIRC, it does not pass-through attributes, something related to const problems), but I am not in the mood to touch Qi/Karma.

djowel:
Good point. Thoughts, @hkaiser ?

Kojoley:
If you want to know my point on Qi/Karma `make_attribute` - it should be merged into `transform_attribute`, it will solve symmetric transformation (possible, but X3 had some) problems.

Kojoley:
Ping?

Kojoley:
@hkaiser seems to be active on GitHub but does not answer here :-(

hkaiser:
@Kojoley I don't know anything about X3, so I'd rather not say anything.

Kojoley:
> @Kojoley I don't know anything about X3, so I'd rather not say anything.

There is #464 for Qi/Karma with 'backported' changes.

PR #472, opened 2019-02-26
X3: Improve situation with move only types

### LINKED ISSUE #465: X3: container element copying bug
Another case of copying AST elements, I have 100+ line template backtrace for this (making minimal example would take some time) but I think that pointing to the code is enough:

`boost/spirit/home/x3/support/traits/container_traits.hpp`

```cpp
    template <typename Container, typename Enable = void>
    struct append_container
    {

    /* [...] */

        template <typename Iterator>
        static void insert(Container& c, Iterator first, Iterator last, mpl::false_)
        {
            c.insert(c.end(), first, last); // attemps to copy into c elements from range [first, last)
        }
    }
```

Kojoley:
Closes #465

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/container_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/container_traits.hpp
@@ -122,7 +122,7 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
         template <typename T>
         static bool call(Container& c, T&& val)
         {
-            c.insert(c.end(), std::move(val));
+            c.insert(c.end(), static_cast<T&&>(val));
             return true;
         }
     };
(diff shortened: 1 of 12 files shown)


djowel:
Nice, as usual!

include/boost/spirit/home/x3/support/traits/container_traits.hpp
```diff
@@ -122,15 +122,15 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
...
+            c.insert(c.end(), static_cast<T&&>(val));
             return true;
         }
     };
 
     template <typename Container, typename T>
     inline bool push_back(Container& c, T&& val)
     {
-        return push_back_container<Container>::call(c, std::move(val));
+        return push_back_container<Container>::call(c, static_cast<T&&>(val));
```
Xeverous: What's the reason behind such changes? Reducing use of standard library (compile time)?

Kojoley: Removed surprising move of lvalue. And yes, `std::forward` is not zero cost in terms of compile time (also a surprising thing).

PR #485, opened 2019-03-18
Correctly handle floating-point "equality" in qi/numeric/real.hpp

gpantwrs:


Kojoley:
I think neither of methods is right. The epsilon in every particular case is a problem specific value. Either the literal parser should be customizable with a comparator, or probably it should just not exist (or fine as-is), because you can do the comparison in a semantic action.

djowel:
Not sure. Maybe make it customizable; or maybe at least make epsilon customizable with a default.

Kojoley:
@gpantwrs would you like to update the PR with a generic solution to the problem?

djowel:
I'd welcome a generic solution to this. For now, I'll close this and leave the code as-is.

PR #489, opened 2019-04-16
Explicitly delete all non implemented assignment operators.

Romain-Geissler-1A:
Commit (almost) automatically generated with the following sed command:
sed -i -ne '1h;1!H;${g;s|// silence MSVC warning C4512: assignment operator could not be generated\n\([[:blank:]]*\)\([^\n]\+\);\n|// silence MSVC warning C4512: assignment operator could not be generated\n\1BOOST_DELETED_FUNCTION(\2);\n|g;p}' $(git ls-files)

Then all files in the x3 subfolder were reverted to HEAD, and manually
updated to use " = delete" instead of BOOST_DELETED_FUNCTION.

### DIFF
--- a/include/boost/spirit/home/karma/auto/auto.hpp
+++ b/include/boost/spirit/home/karma/auto/auto.hpp
@@ -146,9 +146,8 @@ namespace boost { namespace spirit { namespace karma
         T t_;
         generator_impl_type generator_;
 
-    private:
         // silence MSVC warning C4512: assignment operator could not be generated
-        lit_auto_generator& operator= (lit_auto_generator const&);
+        BOOST_DELETED_FUNCTION(lit_auto_generator& operator= (lit_auto_generator const&));
     };
 
     ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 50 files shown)


Kojoley:
> Then all files in the x3 subfolder were reverted to HEAD, and manually
> updated to use " = default" instead of BOOST_DELETED_FUNCTION.

It should be ` = delete;`, is not it?

Romain-Geissler-1A:
> > Then all files in the x3 subfolder were reverted to HEAD, and manually
> > updated to use " = default" instead of BOOST_DELETED_FUNCTION.
> 
> It should be ` = delete;`, is not it?

Yes I just saw the mistake as well, and fixed both the diff + the commit message + the PR description.

include/boost/spirit/home/karma/auto/auto.hpp
```diff
@@ -148,7 +148,7 @@ namespace boost { namespace spirit { namespace karma
 
     private:
```
Kojoley: Those should be removed.

Romain-Geissler-1A: Done

Kojoley:
Thanks!

PR #494, opened 2019-04-22
Replace a couple of dead links.

tinko92:
Hi, this PR fixes a couple for broken links. I did my best to ensure that they refer to the same information as the original links (using the web archive). Some modified files (like classic/phoenix/doc/preface.html) state that they are generated at the top but I failed to find the corresponding source files for their generation in the repository, so I modified them anyway.


djowel:
Nice one! Thank you very much. I'm glad you were able to find the corrected links.

PR #507, opened 2019-05-06
X3: Simplify string_traits

Kojoley:
We probably may completely remove `string_traits.hpp` by rewriting things further.


djowel:
Wonderful!

Issue #520, opened 2019-07-07
any_char parser accepts chars outside encoding


sehe:
The following assert fails:

```c++
#include <boost/spirit/home/x3.hpp>

int main() {
    namespace x3 = boost::spirit::x3;

    char const* input = "\x80";
    assert(!x3::parse(input, input+1, x3::ascii::char_));
}
```

The assert should pass because `char_encoding::ascii::test` disallows non-7bit ASCII, so `"\x80"` is outside the encoding.

The corresponding Qi test case DOES pass:

```c++
#include <boost/spirit/include/qi.hpp>

int main() {
    namespace qi = boost::spirit::qi;

    char const* input = "\x80";
    assert(!qi::parse(input, input+1, qi::ascii::char_));
}
```

As noticed on stackoverflow https://stackoverflow.com/questions/56916592/spirit-x3-how-to-fail-parse-on-non-ascii-input

sehe:
Created a merge request the most logical fix, https://github.com/boostorg/spirit/pull/521

Should not break any tests

djowel:
merged.

PR #521, opened 2019-07-07
Fix for #520 any_char accepts outside encoding

sehe:
The following assert fails:

    int main() {

        char const* input = "\x80";
        assert(!x3::parse(input, input+1, x3::ascii::char_));
    }

The assert should pass because char_encoding::ascii::test disallows
non-7bit ASCII, so "\x80" is outside the encoding.

The corresponding Qi test case DOES pass:

    int main() {

        char const* input = "\x80";
        assert(!qi::parse(input, input+1, qi::ascii::char_));
    }

This simple fix requires the encoding's `ischar` to return true always,
making the behavior as specified and consistent with Qi and docs.

### DIFF
--- a/include/boost/spirit/home/x3/char/any_char.hpp
+++ b/include/boost/spirit/home/x3/char/any_char.hpp
@@ -24,7 +24,7 @@ namespace boost { namespace spirit { namespace x3
         template <typename Char, typename Context>
         bool test(Char ch_, Context const&) const
         {
-            return ((sizeof(Char) <= sizeof(char_type)) || encoding::ischar(ch_));
+            return ((sizeof(Char) <= sizeof(char_type)) && encoding::ischar(ch_));
         }
 
         template <typename Char>


djowel:
Environment: APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2015, ADDRMDL=64, TOOLSET=msvc-14.0 seems to be failing. Hrmmm...

sehe:
@djowel Looks unrelated. Only on that version of the compiler, is a compilation error anyways.

The same error occurs on other recent builds like https://ci.appveyor.com/project/Kojoley/spirit-jod4k/builds/25691801 (PR 511)

djowel:
Yeah, looks unrelated. I'll go merge this one.

Kojoley:
Did you intentionally make `ascii::char_` always failing for iterators with bigger types than `char`, for example with UTF-32:

```cpp
#include <string>
#include <iostream>

int main()
{
    std::u32string s = U"ascii";
    bool r = parse(s.begin(), s.end(), +boost::spirit::x3::ascii::char_);
    std::cout << std::boolalpha << r << '\n';
}
```

It is a legitimate usage, is not it?

djowel:
> Did you intentionally make `ascii::char_` always failing for iterators with bigger types than `char`, for example with UTF-32:

Not sure. But, when in doubt, check what Qi code does.

Issue #533, opened 2019-09-09
boost::spirit::x3::traits::move_to: needs boost::optional<Source> to work with -optional


duzy:
For a `def` like (https://github.com/extbit/mold/blob/aa0fe70e75d2dfb56915b8307728448d376660a9/include/mold/domain/tildache/parser_def.hpp#L151):
```c++
    auto const tild_see_section_def =
      tild_see_case
      >> +tild_expr_case
      >> -tild_else_case
      >> omit[sk[lit("{{") >> '~' >> "end" >> -lit("see") >> "}}"]]
      ;
```
Where my ast is `tild_else_case = boost::optional<node_list>`, the `taits::move_to` is not working:
```c++
    template <typename Source, typename Dest>
    inline void move_to(Source&& src, Dest& dest)
    {
        detail::move_to(std::move(src), dest
          , typename attribute_category<Dest>::type());
    }
```

Reason: when `Source` is boost::optional, `detail::move_to` has no viable working version. (In my case `node_list` is a `container_attribute`.)

This specialization fixes the issue for me (https://github.com/extbit/mold/blob/aa0fe70e75d2dfb56915b8307728448d376660a9/include/mold/domain/tildache/ast.hpp#L12):

```c++
namespace boost { namespace spirit { namespace x3 { namespace traits
{
  template <typename Source, typename Dest>
  inline void move_to(boost::optional<Source>&& src, Dest& dest)
  {
    // For tild_else_case = boost::optional<node_list>
    if (src) move_to(*src, dest);
  }
}}}} // boost::spirit::x3::traits
```

But a better fix should be added to `boost/spirit/home/x3/support/traits/move_to.hpp`!

djowel:
Could you do a PR for this, including a test that exhibits the issue?

duzy:
Hi @djowel , I think this trivial test should reveal the issue:
```c++
// This test exhibits the issue of parsing to boost::optional<ast::foo> node.
// It works well in previous version of spirit::x3.
//
// Duzy Chan, 2019-09-11, ExtBit Limited

#include <boost/spirit/home/x3.hpp>
#include <boost/fusion/include/adapt_struct.hpp>

namespace extbit
{
  namespace ast
  {
    struct foo
    {
      int bar;
      int baz;
    };

    using foo_opt = boost::optional<foo>;

    struct top
    {
      foo a;
      foo_opt b;
      foo c;
    };
  }
}

[...]

PR #534, opened 2019-09-11
Add handling with `optional<Source>` for `boost::spirit::x3::traits::…

duzy:
…move_to`,

so that optional ast node will work. See also test/x3/optional_ast_node.cpp for
the case.

Addressed at https://github.com/boostorg/spirit/issues/533

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/move_to.hpp
+++ b/include/boost/spirit/home/x3/support/traits/move_to.hpp
@@ -189,6 +189,13 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
     }
 
     template <typename Source, typename Dest>
+    inline void move_to(boost::optional<Source>&& src, Dest& dest)
+    {
+        if (src) detail::move_to(std::move(*src), dest
+          , typename attribute_category<Dest>::type());
+    }
+
+    template <typename Source, typename Dest>
     inline void move_to(Source&& src, Dest& dest)
     {
         detail::move_to(std::move(src), dest
--- a/meta/libraries.json
+++ b/meta/libraries.json
@@ -6,7 +6,7 @@
         "Hartmut Kaiser",
         "Dan Nuffer"
     ],
-    "description": "LL parser framework represents parsers directly as EBNF grammars in inlined C++. ",
+    "description": "LL parser framework represents parsers directly as EBNF grammars in inlined C++.",
     "category": [
         "Parsing",
         "String"
--- a/meta/libraries.json
+++ b/meta/libraries.json
@@ -6,7 +6,7 @@
         "Hartmut Kaiser",
         "Dan Nuffer"
     ],
-    "description": "LL parser framework represents parsers directly as EBNF grammars in inlined C++.",
+    "description": "LL parser framework represents parsers directly as EBNF grammars in inlined C++. ",
     "category": [
         "Parsing",
         "String"
--- a/test/x3/optional_ast_node.cpp
+++ b/test/x3/optional_ast_node.cpp
@@ -1,0 +1,66 @@
+/*=============================================================================
+    Copyright (c) 2001-2015 Joel de Guzman
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Duzy Chan:
+    This test addresses the usage of boost::optional<foo> as an ast node.
+=============================================================================*/
+#include <boost/detail/lightweight_test.hpp>
+#include <boost/spirit/home/x3.hpp>
+#include <boost/fusion/adapted/struct.hpp>
+#include <boost/fusion/include/vector.hpp>
+
+#include <iostream>
+#include "test.hpp"
+#include "utils.hpp"
+
+struct twoints
+{
+    int a;
+    int b;
+};
+
+struct adata
+{
+  twoints a;
+  boost::optional<twoints> b;
+  twoints c;
+};
+
+BOOST_FUSION_ADAPT_STRUCT(twoints, a, b)
+BOOST_FUSION_ADAPT_STRUCT(adata, a, b, c)
+
+int
+main()
+{
+    {
+      // Duzy Chan: This case addresses the usage of boost::optional<foo>
+      // as an ast node. Which should actually test the ability of
+      // boost::spirit::x3::traits::move_to to handle with optional source
+      // value.
+      boost::spirit::x3::rule<class twoints, adata> top = "top";
+      boost::spirit::x3::rule<class twoints, boost::optional<twoints>>
+        twoints = "twoints";
+     
+      using boost::spirit::x3::int_;
+      auto const top_def = twoints >> ',' >> -twoints >> ',' >> twoints;
+      auto const twoints_def = int_ >> int_;
+
+      BOOST_SPIRIT_DEFINE(top, twoints);
+
+      twoints a, b;
+      BOOST_TEST((test_attr("1 2,3 4,5 6", top, a)));
+      BOOST_TEST((a.a.a == 1 && a.a.b == 2));
+      BOOST_TEST((a.b && a.b->a == 3 && a.b->b == 4));
+      BOOST_TEST((a.c.a == 5 && a.c.b == 6));
+
+      BOOST_TEST((test_attr("1 2,,5 6", top), b));
+      BOOST_TEST((b.a.a == 1 && b.a.b == 2));
+      BOOST_TEST((!a.b));
+      BOOST_TEST((b.c.a == 5 && b.c.b == 6));
+    }
+
+    return boost::report_errors();
+}


djowel:
Wonderful. Thanks!

Kojoley:
@djowel Why you let this get in? In my opinion it must not compile and it was, now this allows to shoot the foot off. It just happens that in the example/test definition of 
`twoints` always sets the value, but `optional` tells that it may not, and in this case reading `adata::c` or copying the whole `adata` structure is UB.

Kojoley:
Well, and the added test does not compile, it was not added to `Jamfile` to be run.

djowel:
@Kojoley I'm probably not understanding the deeper issues here, so please go ahead with your revert.

duzy:
@Kojoley I just responded from the thread https://github.com/boostorg/spirit/issues/533#issuecomment-640331669 . I'm not sure what's real problem is by adding `boost:optional` RHS for `move_to` op. If you can paste the code to show the case, I'm very interested to review it for the problem.

Kojoley:
> I'm not sure what's real problem is by adding `boost:optional` LHS for `move_to` op. 

Here is the problem:

```cpp
#include <boost/optional.hpp>
#include <iostream>

const x3::rule<class foo_r, boost::optional<int>> foo = "foo";
auto const foo_def = x3::eps;
BOOST_SPIRIT_DEFINE(foo)

int main()
{
    std::string s = "";
    int i;
    if (parse(s.begin(), s.end(), foo, i))
        std::cout << "success: " << i << '\n';
    else
        std::cout << "failed";
}
```

After this PR it compiles and it is invoking UB.

duzy:
> > I'm not sure what's real problem is by adding `boost:optional` LHS for `move_to` op.
> 
> Here is the problem:
> 
> ```c++
> #include <boost/spirit/home/x3.hpp>
> #include <boost/optional.hpp>
> #include <iostream>
> 
> namespace x3 = boost::spirit::x3;
> 
> const x3::rule<class foo_r, boost::optional<int>> foo = "foo";
> auto const foo_def = x3::eps;
> BOOST_SPIRIT_DEFINE(foo)
> 
> int main()
> {
>     std::string s = "";
>     int i;
>     if (parse(s.begin(), s.end(), foo, i))
>         std::cout << "success: " << i << '\n';
>     else
>         std::cout << "failed";
> }
> ```
> 
> After this PR it compiles and it is invoking UB.

If I understand it correctly, the `eps` parses nothing (zero data). The grammar rule `foo` in this case should be undefined behavior:

    * eps is not a `int` parser, nor a `optional<int>` parser
    * the `foo` grammar is not a `int` parser, but a `optional<int>` parser

So in your example,  `parse(s.begin(), s.end(), foo, i)` should be undefined (I'm not sure if it's what you mean by *UB*?).

What if you try with this:
```c++
#include <boost/optional.hpp>
#include <iostream>

const x3::rule<class foo_r, boost::optional<int>> foo = "foo";
auto const foo_def = x3::eps;
BOOST_SPIRIT_DEFINE(foo)

int main()
{
    std::string s = "";
    boost::optional<int> i;
    if (parse(s.begin(), s.end(), foo, i))
        std::cout << "success: " << i << '\n';
    else
        std::cout << "failed";
}
```

Noted that `foo` is for parsing `optional<int>` not `int`, what do you expect with it for a `int`??

Kojoley:
You are focusing on the wrong thing, there is no problem with `eps`, you can replace it with `-int_` and it will be the same thing, I used `eps` to show the extreme absurdity of the situation.

If you need an analogy, what you have done in the PR is similarly to allow implicit pointer to reference conversion.

duzy:
> You are focusing on the wrong thing, there is no problem with `eps`, you can replace it with `-int_` and it will be the same thing, I used `eps` to show the extreme absurdity of the situation.
> 
> If you need an analogy, what you have done in the PR is similarly to allow implicit pointer to reference conversion.

Are you saying that the PR makes this undefined code passed compilation?
```c++
int main()
{
    std::string s = "";
    int i;
    if (parse(s.begin(), s.end(), foo, i))
        std::cout << "success: " << i << '\n';
    else
        std::cout << "failed";
}
```

Since the `foo` parser is expecting to work with `optional<int>`, when parse it for `int`, it should complain for errors. If the PR makes it pass compilation, it should need a new fix or better fix instead of reverting (because it disabled the `optional<bar>` field issue again).

A proper try might be like this:
```c++
    template <typename Source, typename Dest>
    inline void move_to(boost::optional<Source>&& src, boost::optional<Dest>&& dest)
    {
        if (src) detail::move_to(std::move(*src), dest
          , typename attribute_category<Dest>::type());
    }
```
To replace this:
```c++
    template <typename Source, typename Dest>
    inline void move_to(boost::optional<Source>&& src, Dest&& dest)
    {
        if (src) detail::move_to(std::move(*src), dest
          , typename attribute_category<Dest>::type());
    }
```
So that a `optional<int>` parser won't work with `int` again (by mistake).

Anyway it's up to you, I'm so glad that you point it out.

Kojoley:
> Are you saying that the PR makes this undefined code passed compilation?

Yes, in every my message, and you are saying about it:

https://github.com/boostorg/spirit/issues/533#issuecomment-640674172:
> No worry about the safety of `move_to`, we know what it's doing and when when we should use it.

---

> A proper try might be like this:
> ```cpp
>     template <typename Source, typename Dest>
>     inline void move_to(boost::optional<Source>&& src, boost::optional<Dest>&& dest)
>     {
>         if (src) detail::move_to(std::move(*src), dest
>           , typename attribute_category<Dest>::type());
>     }
> ```

There is no need in it, see example in https://github.com/boostorg/spirit/issues/533#issuecomment-640540916, the https://github.com/boostorg/spirit/blob/e01e9fbfdaf2c16332c4da89e9b563ee70ea95e9/include/boost/spirit/home/x3/support/traits/move_to.hpp#L155-L160 since beginning.

duzy:
> > Are you saying that the PR makes this undefined code passed compilation?
> 
> Yes, in every my message, and you are saying about it:
> 
> [#533 (comment)](https://github.com/boostorg/spirit/issues/533#issuecomment-640674172):
> 
> > No worry about the safety of `move_to`, we know what it's doing and when when we should use it.
> 
> > A proper try might be like this:
> > ```c++
> >     template <typename Source, typename Dest>
> >     inline void move_to(boost::optional<Source>&& src, boost::optional<Dest>&& dest)
> >     {
> >         if (src) detail::move_to(std::move(*src), dest
> >           , typename attribute_category<Dest>::type());
> >     }
> > ```
> 
> There is no need in it, see example in [#533 (comment)](https://github.com/boostorg/spirit/issues/533#issuecomment-640540916), the
> 
> https://github.com/boostorg/spirit/blob/e01e9fbfdaf2c16332c4da89e9b563ee70ea95e9/include/boost/spirit/home/x3/support/traits/move_to.hpp#L155-L160
> 
> handles it since the beginning.

Thanks @Kojoley , it's up to you to decide how to fix it. But notice that `move_to(Source&& src, Dest& dest, optional_attribute)` require a extra attribute argument, which means it requires extra settings from client code. In your point of view, it should be done via something like `transform_attribute`, but in our case or for complex/large AST, we need it exactly to support that `optional<bar>` struct member to be passed.

But when I said "no worry about the safety of move_to", there's no offense. I mean it for real. Spirit X3 is perfect for us to quickly prototype a grammar for parsing or for quick utility. Normally we expect errors to be raised as soon as possible. It's obviously the best time to complain about errors at compile time. Or we prefer to see it crash as soon as possible if there's errors. When we change `move_to` for the fix, we know that it's for our case specifically, so we know what's happening (rather than using an `optional<int>` parser for an `int`).

The complexity of spirit::x3 is actually driven us to write language parsing code in C++ from scratch. For large AST and complex semantics, spirit::x3 is not a good choice, it mixes everything and difficult error tracking spent us more time than self-written language parsing.

Thant's why I would suggest you to simply spirit::x3 as much as possible, instead of introducing too many settings like `transform_attribute`. It's really not a good option for large AST tree.

Kojoley:
It seems that you still not understand the issue. The point why `optional<int>` should not parse into `int` is because `optional<int>` has no implicit conversion to `int`, i.e.:
```cpp
boost::optional<int> foo();
int i = foo();
```
will not compile, because authors of `optional` decided to not allow it, and for good reason. If you knew upfront that the `optional<T>` will always has a value in it, you simply do not need `optional`, and could just replace the type with the plain `T`, but if you did not, it means that there is something broken with your types or grammar.

I am open to simplifications in the library, if you have a justified examples I am happy to try to make them working.

duzy:
> It seems that you still not understand the issue. The point why `optional<int>` should not parse into `int` is because `optional<int>` has no implicit conversion to `int`, i.e.:
> 
> ```c++
> boost::optional<int> foo();
> int i = foo();
> ```
> 
> will not compile, because authors of `optional` decided to not allow it, and for good reason. If you knew upfront that the `optional<T>` will always has a value in it, you simply do not need `optional`, and could just replace the type with the plain `T`, but if you did not, it means that there is something broken with your types or grammar.

You're right, `optional` is actually more like a special *pointer* or *reference*. I think I was actually talking about `variant` when I said 'initialized into the first template type' (see https://github.com/duzy/mold/blob/caf55317662573380333c56105a8e0de08871f13/include/mold/domain/tildache/ast.hpp#L62). Because the example of the *mold* project was actually written in around 2016 for a quick utility. It's been years, so I have to think and recall the memory. I can only think of more when going back to and reading the code while reading your words to find out what's happening.

The truth in our case was that when we're using `optional`, we really want it to be an *option* to be absent. This is exactly the semantic of the language ast of `optional`. So the usage of `optional` is correct for our case.

But if the PR is making an `optional<int>` parser generating for `int`, it's obviously best to restrict it at compile time (it's properly depending on whether if an `int` is movable into an `optional<int>` (the && semantic)). The semantics of `-foo` for generating `using foo = optional<bar>` (in my case as a struct member) is indeed a natural needs for simplification without the needs for extra settings like `transform_attribute` or other specializations/overriding. So if will be great if you can come up with a fix to consider the both.

I think narrowing down the `Dest` type for `move_to` could be a solution to disable the undefined-behavior code at compile time.

> 
> I am open to simplifications in the library, if you have a justified examples I am happy to try to make them working.

I wish I could try something for the moment, but it's been changed a lot of my situation. I think my last change related to this should be https://github.com/extbit/mold/commit/1d2978d588a86e0f70855bbc8fe5e1ced50596af, and the example test should address the failure issue (if not applying this PR):
```c++
namespace extbit
{
  namespace ast
  {
    struct bar
    {
      int baz1;
      int baz2;
    };

    using foo = boost::optional<bar>;

    struct top
    {
      bar a;
      foo b;
      bar c;
    };
  }
}

BOOST_FUSION_ADAPT_STRUCT(extbit::ast::top, a, b, c)
BOOST_FUSION_ADAPT_STRUCT(extbit::ast::bar, baz1, baz2)

namespace extbit
{
  namespace parser
  {

    const x3::rule<class top, ast::top> top = "top";
    const x3::rule<class foo, ast::bar> foo = "foo";
    using x3::int_;

    auto const top_def = foo >> -foo >> foo;
    auto const foo_def = int_ >> int_;

    BOOST_SPIRIT_DEFINE(top, foo);
  }
}
```

Kojoley:
Once again, your example is ill-formed (I showed a fix in https://github.com/boostorg/spirit/issues/533#issuecomment-640246813), in `top_def = foo >> -foo >> foo;` the third `foo` parses into `bar`, it asks to parse `optional<bar>` into `bar`!

duzy:
> Once again, your example is ill-formed (I showed a fix in [#533 (comment)](https://github.com/boostorg/spirit/issues/533#issuecomment-640246813)), in `top_def = foo >> -foo >> foo;` the third `foo` parses into `bar`, it asks to parse `optional<bar>` into `bar`!

Okay, please go with your fix on that. It was spelling error not unrelated to the real issue. Last time I wrote this example was just quickly extracting the case from _mold_ to show the issue.

Issue #537, opened 2019-09-21
MSVC C4127 warning workarounds vs BOOST_CONSTEXPR


rubenvb:
There are several places in Spirit that trigger the C4127: conditional expression constant warning.

Essentially these should be `if constexpr` but I see the warning being disabled using pragma's.

I'll make a PR adding a similar pragma push/pop around the occurences I see building my project.

It would better IMHO to add `BOOST_CONTEXPR` after the relevant `if`s, as we can't just rely on `if constexpr` in a C++14 library. I'll make a PR for this too.

rubenvb:
Adding BOOST_CONSTEXPR doesn't work due to C++11/14 not having `if constexpr`, so never mind that. The warning (it was only a single occurence I see in my code) can still be silenced locally.

PR #538, opened 2019-09-21
Disable warning about constant conditional expression.

rubenvb:
Fixes issue #537, part 1.

### DIFF
--- a/include/boost/spirit/home/x3/char/detail/cast_char.hpp
+++ b/include/boost/spirit/home/x3/char/detail/cast_char.hpp
@@ -8,6 +8,8 @@
 #if !defined(BOOST_SPIRIT_X3_CAST_CHAR_NOVEMBER_10_2006_0907AM)
 #define BOOST_SPIRIT_X3_CAST_CHAR_NOVEMBER_10_2006_0907AM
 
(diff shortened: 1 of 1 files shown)


include/boost/spirit/home/x3/char/detail/cast_char.hpp
```diff
@@ -22,7 +22,10 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
     // The trick is to cast to an unsigned version of the source char first
     // before casting to the target. {P.S. Don't worry about the code, the
     // optimizer will optimize the if-else branches}
-
+#if BOOST_WORKAROUND(BOOST_MSVC, >= 1400)
```
Kojoley: Please, either include boost workaround header or do not use it. Also, boost is not tested/supported under msvc < 16 for a while already, so you can just check for msvc compiler, i.e. just `ifdef _MSC_VER `.

rubenvb: I included the relevant header. I kept BOOST_WORKAROUND because that's how this kind of thing is done in all other Spirit headers.

Kojoley:
Superseded by #541

PR #539, opened 2019-09-21
Sprinkle BOOST_CONSTEXPR on constexpr ifs.

rubenvb:
These probably don't improve performance, but should inhibit any future compiler warnings about constant conditional expressions (outside of the already worked around MSVC's C4127).

Part 2 of fixes for #537


rubenvb:
Right. Sorry for the brainfart. Never mind about this then.

PR #541, opened 2019-09-29
Fix MSVC C4127 warnings

### LINKED ISSUE #537: MSVC C4127 warning workarounds vs BOOST_CONSTEXPR

Kojoley:
Closes #537


djowel:
Looks good to me.

PR #543, opened 2019-10-09
Fixed undefined behaviour when calling functions from <cctype>

wmamrak:
It is required to cast ch from signed char to unsigned char, see:
https://en.cppreference.com/w/cpp/string/byte/islower#Notes

### DIFF
--- a/include/boost/spirit/home/support/char_encoding/standard.hpp
+++ b/include/boost/spirit/home/support/char_encoding/standard.hpp
@@ -41,61 +41,61 @@ namespace boost { namespace spirit { namespace char_encoding
         static bool
         isalnum(int ch)
         {
-            return std::isalnum(ch) ? true : false;
+            return std::isalnum((unsigned char)ch) ? true : false;
         }
 
         static bool
         isalpha(int ch)
         {
-            return std::isalpha(ch) ? true : false;
+            return std::isalpha((unsigned char)ch) ? true : false;
         }
 
         static bool
         isdigit(int ch)
         {
-            return std::isdigit(ch) ? true : false;
+            return std::isdigit((unsigned char)ch) ? true : false;
         }
 
         static bool
         isxdigit(int ch)
         {
-            return std::isxdigit(ch) ? true : false;
+            return std::isxdigit((unsigned char)ch) ? true : false;
         }
 
         static bool
         iscntrl(int ch)
         {
-            return std::iscntrl(ch) ? true : false;
+            return std::iscntrl((unsigned char)ch) ? true : false;
         }
 
         static bool
         isgraph(int ch)
         {
-            return std::isgraph(ch) ? true : false;
+            return std::isgraph((unsigned char)ch) ? true : false;
         }
 
         static bool
         islower(int ch)
         {
-            return std::islower(ch) ? true : false;
+            return std::islower((unsigned char)ch) ? true : false;
         }
 
         static bool
         isprint(int ch)
         {
-            return std::isprint(ch) ? true : false;
+            return std::isprint((unsigned char)ch) ? true : false;
         }
 
         static bool
         ispunct(int ch)
         {
-            return std::ispunct(ch) ? true : false;
+            return std::ispunct((unsigned char)ch) ? true : false;
         }
 
         static bool
         isspace(int ch)
         {
-            return std::isspace(ch) ? true : false;
+            return std::isspace((unsigned char)ch) ? true : false;
         }
 
         static bool
@@ -107,7 +107,7 @@ namespace boost { namespace spirit { namespace char_encoding
         static bool
         isupper(int ch)
         {
-            return std::isupper(ch) ? true : false;
+            return std::isupper((unsigned char)ch) ? true : false;
         }
 
     ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


wmamrak:
Fixes assertion failure of ch being < 0 on Visual Studio 2019 for the following test case:

```

int main() {
    std::string txt = "ą";

    auto b = txt.begin();
    auto e = txt.end();

    parse(b, e, no_case['a']);

    return 0;
}
```

Important note: the file must be saved with UTF-8 encoding without BOM.

djowel:
I don't think that is the issue, or the solution here (including the link you gave). The caller already receives an int and passes the int. It's probably somewhere outside the char_encoding traits. I'll have a look.

Side note: std::isxxx(ch) ? true : false; is odd, style-wise. I don't know when that started. It should just be std::isxxx(ch), without the conditional.

djowel:
> Also, correct version would be std::isxxx(ch) != 0

Perhaps that's the reason for the std::isxxx(ch) ? true : false;

... I'll look into this as soon as I can.

djowel:
> In my example, the code point "ą" consists in UTF-8 encoding of 2 code units: 0xc4 and 0x85.
> The no_case_directive::parse calls char_parser::parse

Aside: Oh and BTW, you shouldn't pass utf8 as-is. Instead use a unicode iterator (e.g. the ones in boost regex) and the unicode encoding. no_case does not make sense for utf8 encoded string, unlike the unicode encoding.

wmamrak:
I assumed X3's behaviour did not change since version 2. In version 2, if I am not mistaken, literal_char parser transformed the argument of no_case ("a" in my example) to its lower and upper versions, and then made the comparison to characters in the input string, whatever that input string encoding was. It passed all of the exhaustive tests, so I assumed there are no reasons to add overhead of using utf8-aware iterators.

wmamrak:
To sum the above up, as long as none, or only one character from input string and no_case's argument is utf8-encoded, matching should work.

djowel:
> I assumed X3's behaviour did not change since version 2. In version 2, if I am not mistaken, literal_char parser transformed the argument of no_case ("a" in my example) to its lower and upper versions, and then made the comparison to characters in the input string, whatever that input string encoding was. It passed all of the exhaustive tests, so I assumed there are no reasons to add overhead of using utf8-aware iterators.

Yes, that is correct. However, Qi nor X3 does not expand utf8 (which can comprise multiple bytes for a single code point) automatically. What you want for utf8 is the unicode encoding, but you have to do the utf8 to ucs4 decoding before hand using an iterator adaptor such as the ones in boost.regex.

(Note: no_case in that case makes use of http://bit.ly/33m8lD8)

See https://github.com/cierelabs/json_spirit for an example on how to correctly parse utf8.

djowel:
To sum the above, if you need to parse utf8 correctly, you need to use the unicode encoding and the utf8-aware iterators. Having said that, this issue 543 is valid and should be fixed.

djowel:
> To sum the above, if you need to parse utf8 correctly, you need to use the unicode encoding and the utf8-aware iterators. Having said that, this issue 543 is valid and should be fixed.

OK, let me relax that a little bit. After syncing my thoughts, I realised that you only need the utf8-aware iterators IF you are using the character classification parsers, including the no_case directive. Why? Because otherwise, using only the standard encoding, you cannot convert unicode code-points such as this example: 'Ž' -> 'ž'

In the json parser I gave a link to above, the unicode iterators are actually just needed for converting json escaped strings, but all else simply use the ascii encoding.

wmamrak:
> on utf8 inputs, you only need the utf8-aware iterators IF you are using the character classification parsers, including the no_case directive

Yes, but for the no_case directive this is even less strict, i.e. utf8-aware iterators are needed only IF input string and no_case's argument BOTH contain utf8 characters.

wmamrak:
Upon further investigation I found a related issue. Assume I want to parse a utf8 input with an ascii::space skipper. 
```

int main() {
    std::string txt = "ą";

    auto b = txt.begin();
    auto e = txt.end();

    phrase_parse(b, e, no_case['a'], ascii::space);

    return 0;
}
```
Since it is a classifier, as you said, parsing will fail with an assertion in
boost/spirit/home/support/char_encoding/ascii.hpp
in the function ascii::isspace (line 264 atm).
Yet again it is a utf8 code unit > 127, stored as a char in a std::string (hence < 0), then casted to int.
One can solve this in a few ways:
1. write a custom parser which does not call boost::spirit::char_encoding::ascii::isspace or any function from cctype
2. use literal_char<char_encoding::ascii, unused_type> skipper = ' '; which is awful
3. introduce a cast in boost/spirit/home/support/char_encoding/ascii.hpp similarly to what I proposed in this PR.
4. do something more intrusive that 3
5. use unicode::space . This will bloat your executable significantly and introduce additional complexity thus slowing it down, for no reason really.

wmamrak:
Related: 22040cb3b22c589be66bf610b258a8da44d4613d

djowel:
Fixed in https://github.com/boostorg/spirit/commit/f10cf8a80f31926e73d180b81bcb721db4b7cfd8
(with safety support in https://github.com/boostorg/spirit/commit/5a217a5e033d34b016fef41ee95e311bc1a24140)

wmamrak:
The core of the issue remained, though:
```

int main() {
    std::string txt = "ą";

    assert(!std::isdigit((unsigned char)txt[0]));
    assert(!std::isdigit((unsigned char)txt[1]));

    auto b = txt.begin();
    auto e = txt.end();

    assert(!parse(b, e, digit));

    return 0;
}
```

djowel:
I'll look into it later. I suggest you post a complete test case that I can add to the test suite. Please make another issue for it.

PR #549, opened 2019-10-18
X3: Fix `parse_rule` definition check

### LINKED ISSUE #548: Regression since BOOST 1_70
Minimal use-case : https://github.com/artemp/spirit-x3-test ( somewhat related to #511) 
This code compiles, links and runs as expected with BOOST <= 1.69 (both GCC and clang)  but results in linker error with BOOST >=1.70. Key features are separate TUs for grammar instantiations and use of semantic actions in aggregate grammar. 

/cc @djowel @Kojoley

Kojoley:
Regressed in #457.

Fixes #548.


djowel:
Nice one!

PR #550, opened 2019-11-04
V2: Use Boost.Endian

Kojoley:
And drop in-house implementation.


djowel:
Wonderful! Thanks @Kojoley !

PR #556, opened 2019-11-23
Rename variables and suppress warnings about hiding global declaratio…

rubenvb:
…ns with local declarations.

This silences some occurrences of C4459 when using VS2019.


djowel:
@rubenvb , I'm sorry if this is very late, but can you resolve the conflicts, so we can merge this in?

rubenvb:
I rebased the original changes onto the current boostorg:develop branch.

djowel:
Thanks! merged.

PR #557, opened 2019-11-23
Make conversions in move_to explicit.

rubenvb:
This fixes warnings emitted by the VS2019 compiler.


rubenvb:
This was a mistake, please ignore.

PR #561, opened 2019-12-06
Added static assertion for meaningless rule attribute qualifiers

Kojoley:
```cpp
qi::rule<char const*, int()>        // ok
qi::rule<char const*, int const()>  // ok, but strange
qi::rule<char const*, int&()>       // meaningless reference
qi::rule<char const*, int const&()> // meaningless reference

ka::rule<char const*, int()>        // ok
ka::rule<char const*, int const()>  // meaningless const
ka::rule<char const*, int&()>       // meaningless reference
ka::rule<char const*, int const&()> // meaningless const and reference
```

### DIFF
--- a/include/boost/spirit/home/karma/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/karma/nonterminal/rule.hpp
@@ -15,8 +15,6 @@
 #include <boost/config.hpp>
 #include <boost/function.hpp>
 #include <boost/mpl/vector.hpp>
-#include <boost/type_traits/add_const.hpp>
-#include <boost/type_traits/add_reference.hpp>
 #include <boost/type_traits/is_same.hpp>
 
 #include <boost/fusion/include/vector.hpp>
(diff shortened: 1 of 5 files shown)


djowel:
Perhaps I haven't thought this through, but can you explain a bit why we need this?

Kojoley:
Because it helps people find misuses easier? They already get an assertion for those misuses but from the `transform_attribute` and they are harder to diagnose.

Btw, I forgot about X3, it also needs this.

djowel:
> Because it helps people find misuses easier? They already get an assertion for those misuses but from the `transform_attribute` and they are harder to diagnose.
> 
> Btw, I forgot about X3, it also needs this.

Is there a related ticket or use-case that prompted you to have this in?

Kojoley:
https://github.com/mapnik/mapnik/blob/745e393fd694908977d9ab5a38ded3446fed88b3/include/mapnik/json/geometry_generator_grammar.hpp#L89-L115 or https://github.com/mapnik/mapnik/blob/c9da5cbd8e563a6b06b3d72d9a883281d1531962/include/mapnik/json/feature_generator_grammar.hpp where `transform_attribute` specializations were added because of bogus `const&` in the rule, plus I had myself the frustration about constref on attributes personally when just started with Spirit back then.

djowel:
OK, the reason I am asking is that I think we need to add a test case for this.

Kojoley:
False-positive tests are just our current test suite. True-positive tests need to trigger compile time error and that is terribly handled with Boost.Build as every check should be a separate test - what hits test suite time significantly, plus compile-fail tests yield error message to the console without ability to toggle them off https://github.com/boostorg/build/issues/262. I can of course put tests and disable them but I do not see any benefits from doing this, also the static assertions are simple enough and I already tested them manually.

djowel:
Yeah. I think we have a few of those, but I am not sure if they are even tested.

PR #570, opened 2020-03-02
X3: constexprize parsers constructors

wanghan02:
Constexpr constructor and constexpr factory function could save valuable execution time. Especially for local parsers, whose construction time could take up to 40% of total execution time in release build and up to 90% in debug build.

1. Some x3 parsers's constructor could easily be constexpr. 
2. unary_parser and binary_parser and all their derived classes cannot directly be constexpr because we introduced Static Initialization Order Fiasco (SIOF) checking (not constexpr) in their constructors in pull request [229](https://github.com/boostorg/spirit/pull/229). I propose that users could have an option to disable this SIOF checking by defining a macro `BOOST_SPIRIT_X3_DISABLE_SIOF_CHECK`. Because

    1) It does not check release build.
    2) The coding style that triggers SIOF is forbidden in many groups. This checking does not bring any benefit in such groups.

Summary: users could use constexpr parsers (except a few ones, e.g. x3::symbols_parser) by defining a macro `BOOST_SPIRIT_X3_DISABLE_SIOF_CHECK` before including x3 headers. This macro will disable SIOF checking. Unit tests are added.

### DIFF
--- a/include/boost/spirit/home/x3/char/char_class.hpp
+++ b/include/boost/spirit/home/x3/char/char_class.hpp
@@ -71,7 +71,7 @@ namespace boost { namespace spirit { namespace x3
 
 #define BOOST_SPIRIT_X3_CHAR_CLASS(encoding, name)                                 \
     typedef char_class<char_encoding::encoding, name##_tag> name##_type;           \
-    name##_type const name = name##_type();                                        \
+    constexpr name##_type name = name##_type();                                        \
     /***/
 
 #define BOOST_SPIRIT_X3_CHAR_CLASSES(encoding)                                     \
--- a/include/boost/spirit/home/x3/char/char_class.hpp
+++ b/include/boost/spirit/home/x3/char/char_class.hpp
@@ -71,7 +71,7 @@ namespace boost { namespace spirit { namespace x3
 
 #define BOOST_SPIRIT_X3_CHAR_CLASS(encoding, name)                                 \
     typedef char_class<char_encoding::encoding, name##_tag> name##_type;           \
-    constexpr name##_type name = name##_type();                                        \
+    constexpr name##_type name = name##_type();                                    \
     /***/
 
 #define BOOST_SPIRIT_X3_CHAR_CLASSES(encoding)                                     \
--- a/include/boost/spirit/home/x3/operator/alternative.hpp
+++ b/include/boost/spirit/home/x3/operator/alternative.hpp
@@ -42,7 +42,7 @@ namespace boost { namespace spirit { namespace x3
     };
 
     template <typename Left, typename Right>
-    inline constexpr alternative<
+    constexpr alternative<
         typename extension::as_parser<Left>::value_type
       , typename extension::as_parser<Right>::value_type>
     operator|(Left const& left, Right const& right)
--- a/include/boost/spirit/home/x3/operator/alternative.hpp
+++ b/include/boost/spirit/home/x3/operator/alternative.hpp
@@ -18,7 +18,7 @@ namespace boost { namespace spirit { namespace x3
     {
         typedef binary_parser<Left, Right, alternative<Left, Right>> base_type;
 
-        alternative(Left const& left, Right const& right)
+        constexpr alternative(Left const& left, Right const& right)
             : base_type(left, right) {}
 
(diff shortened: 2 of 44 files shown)


wanghan02:
> we just simply need what to be marked as constexpr but it should not be really constexpr function.

Thank you for your review. But I still have doubts about this change request. 

1. Even if we make `what(p)` a `constexpr` function (It may be doable because it's a template function, `constexpr` template function only requires at least 1 possible specialization is `constexpr`), it does not mean its specializations for each derived parsers are `constexpr` functions. The fact is they are not and they cannot have a `constexpr` constructor.

2. This SIOF checking is done by checking if `x3::rule::name` is `0` initialized via calling the constructor of `std::string`. Whether `x3::rule::name` is `0` initialized is almost certainly not a compile time result. While `constexpr` function (or `constexpr` template function specialization) requires that all its content could be evaluated at compile time. Otherwise it's not a `constexpr` (compile time evaluated) function.

In my opinion, the SIOF checking and `constexpr` are mutually exclusive. The pull request doesn't change existing application code. Users who want to choose `constexpr` parsers over SIOF could simply enable the macro.

Kojoley:
I am strongly against introducing a new undocumented combinatory explosive switch. I have opened #571, it should solve both issues in a most convenient way.

djowel:
> I am strongly against introducing a new undocumented combinatory explosive switch. I have opened #571, it should solve both issues in a most convenient way.

I am with @Kojoley on this one. But having said that, the move to constexpr is indeed a step in the right direction!

wanghan02:
I have updated the commits based on #571 . Now it's simply constexpr everywhere.

test/x3/constexpr.cpp
```diff
@@ -0,0 +1,75 @@
+/*=============================================================================
+    Copyright (c) 2015 Joel de Guzman
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================*/
+#include <boost/detail/lightweight_test.hpp>
+
+#define BOOST_SPIRIT_X3_DISABLE_SIOF_CHECK
```
Kojoley: This is left by omission, I guess.

include/boost/spirit/home/x3/char/char_class.hpp
```diff
@@ -71,7 +71,7 @@ namespace boost { namespace spirit { namespace x3
 
 #define BOOST_SPIRIT_X3_CHAR_CLASS(encoding, name)                                 \
     typedef char_class<char_encoding::encoding, name##_tag> name##_type;           \
-    name##_type const name = name##_type();                                        \
+    constexpr name##_type name = name##_type();                                        \
```
Kojoley: Since you have denied edits to your branch, please fix backslashes indentation.

include/boost/spirit/home/x3/operator/alternative.hpp
```diff
@@ -42,7 +42,7 @@ namespace boost { namespace spirit { namespace x3
     };
 
     template <typename Left, typename Right>
-    inline alternative<
+    inline constexpr alternative<
```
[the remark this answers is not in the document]
wanghan02: fixed in new commit [667f90f](https://github.com/boostorg/spirit/pull/570/commits/667f90f03a3c0bee9ce639f7c5cea392220331db)

test/x3/constexpr.cpp
wanghan02: fixed in new commit [b7fc3bf](https://github.com/boostorg/spirit/pull/570/commits/b7fc3bf2eb0b5cfab6a79b9f07ac7930fba131ba)

include/boost/spirit/home/x3/char/char_class.hpp
wanghan02: fixed in new commit [ad88408](https://github.com/boostorg/spirit/pull/570/commits/ad88408b4571ffb0144589c2cefa3f6ae68f4144)

Kojoley:
Thanks @wanghan02!

djowel:
> Thanks @wanghan02!

I second that. Thanks!

PR #571, opened 2020-03-02
X3: Alternate rule placeholder static initialization fiasco check

Kojoley:
The previous check was done in a very intrusive way, it also was not catching
the issue when a parser is not inherited from `unary_parser`/`binary_parser`.


wanghan02:
@Kojoley This solution is great! `constexpr` rules cannot trigger SIOF and normal static rules can trigger it at run time. Thank you!

Issue #529, opened 2019-08-28
karma::real_policies<double>::fraction_part() formats incorrectly for precision >= 16 for fractions whose digits form an integer >= 999999999999998


jvanvuren-tableau:
For the current version of Boost, the Karma reals formatting library uses the following code from karma::real_policies<double>::fraction_part():

        static bool fraction_part (OutputIterator& sink, T n
          , unsigned precision_, unsigned precision)
        {
            ...
            typename remove_const<T>::type digits = 
                (traits::test_zero(n) ? 0 : floor(log10(n))) + 1;
            ...

When the precision is >= 16, and "n" (the fractional part of the number converted to an integer) is >= 999999999999998, the expression floor(log10(n))+1 is returning a number of digits required to represent "n" that is 1 too large. When this happens, the final formatted double is 10 times the correct value.

17 digits of precision is the max supported by doubles (https://en.wikipedia.org/wiki/Double-precision_floating-point_format), so it is likely that clients of the Karma library will try to configure it to use that precision, fufilling one of the preconditions for this issue to occur.

Sample code:
```
#include <iostream>

#include <boost/spirit/include/karma.hpp>

namespace karma = boost::spirit::karma;

struct double_policy : karma::real_policies<double> {

    unsigned int precision(double n) const {
        return 16;
    }
};

int main() {

    typedef karma::real_generator<double, double_policy> double_generator;
    const double_generator double_gen = double_generator();

    double x = 0.0999999999999998;
    std::string generated;
    std::back_insert_iterator<std::string> sink(generated);
    karma::generate(sink, double_gen, x);

    std::cout << generated << std::endl;

    return 0;
}
```
Output:
0.999999999999998

hkaiser:
@jvanvuren-tableau Thanks! This looks like a viable solution for larger `n`. Would you have an idea what this `n` could be? Also, would you mind creating a PR for this. I'd gladly accept it.

jvanvuren-tableau:
I've been doing some more reading. The double standard actually only supports 15.9 digits of precision - see https://en.wikipedia.org/wiki/Floating-point_arithmetic. Therefore for someone to try to use karma for format a double with a precision >= 16 digits is invalid. Which would mean that this issue which I submitted is actually invalid :)

hkaiser:
Ok, that's what I was thinking... Are you ok with closing the issue?

jvanvuren-tableau:
Yes - closing issue.

[...]

PR #586, opened 2020-04-10
Karma: Do not limit precision of real_inserter

### LINKED ISSUE #585: [Question] How to create a real_policy to honor max. representable digits?
Hi!

We would like to construct a policy that generates strings from let's say doubles with the minimally needed amount of digits to uniquely read them back in.
Unfortunately, the "precision" member only controls the number of digits in the fraction part and if you have a long integer part, it will create strings that are
a) unreasonably long (since they cannot be accurately represented [EDIT: sorry, they can but are not needed for uniqueness] when read back in) and
~~b) I am not sure how those additional digits are generated from the internal representation and how they even would make sense~~

I understand that I could overwrite precision with something like this:
`return std::numeric_limits<T>::max_digits10::value - floor(log10(n));`

But I am not sure how that would work with scientific representations and how safe and fast this method would be (related to #529 )

jpfeuffer:
Users should be able to decide how much they want to print.
Especially since this implementation does not care about significant digits, which hinders unique serialization when using fixed notation with leading zeros.

AFAIK digits10 and max_digits10 are about the **significant** digits, not the digits in the fractional part.
And for serialization and a string-float-string round-trip you should actually at least allow max_digits10 instead of digits10.

Closes #585

### DIFF
--- a/include/boost/spirit/home/karma/numeric/detail/real_utils.hpp
+++ b/include/boost/spirit/home/karma/numeric/detail/real_utils.hpp
@@ -84,12 +84,6 @@ namespace boost { namespace spirit { namespace karma
 
             // get correct precision for generated number
             unsigned precision = p.precision(n);
-            if (std::numeric_limits<U>::digits10) 
-            {
-                // limit generated precision to digits10, if defined
-                precision = (std::min)(precision, 
-                    (unsigned)std::numeric_limits<U>::digits10 + 1);
-            }
 
             // allow for ADL to find the correct overloads for log10 et.al.
             using namespace std;
--- a/include/boost/spirit/home/karma/numeric/real_policies.hpp
+++ b/include/boost/spirit/home/karma/numeric/real_policies.hpp
@@ -162,8 +162,7 @@ namespace boost { namespace spirit { namespace karma
         //
         //  Note:     If the trailing_zeros flag is not in effect additional
         //            comments apply. See the comment for the fraction_part()
-        //            function below. Moreover, this precision will be limited
-        //            to the value of std::numeric_limits<T>::digits10 + 1
+        //            function below.
         ///////////////////////////////////////////////////////////////////////
         static unsigned precision(T)
         {


hkaiser:
We could limit the precision to `15` instead if we knew that we're dealing with an IEEE 754 conforming FP implementation (see https://en.wikipedia.org/wiki/Floating-point_arithmetic). However, the standard does not assume FP numbers to be IEEE 754 conforming, thus the only way to detect the available precision is to look at `digits10`. 

We've had some discussion on a related topic here: https://github.com/boostorg/spirit/issues/529.  For the reasons outlined there, I'm not fond of completely lifting the limitation.

@jpfeuffer for your purposes I would suggest you switch to represent FP numbers in hexadecimal format, which is guaranteed to represent all bits of the number.

jpfeuffer:
Ok hmm, I indeed (probably wrongly) assumed IEEE754 conformance. But what are potential drawbacks of allowing more digits (except writing some potentially useless digits)? I am just curious.

I think with the current solution you can not print enough digits for numbers like:
0.007519531251111111376517204263336680014617741107940673828125

Boost::karma will print (mind the leading zeros):
0.0075195312511111

Which, when read back in will give you the equivalent of:
0.007519531251111100100814610414090566337108612060546875

Actually, I don't necessarily need this precision but I thought if I could do it lossless (while keeping it
easily readable in decimal format) then I should give it a try.

hkaiser:
@jpfeuffer please go read the discussion in #529, that explains why we can't do that.

Kojoley:
@hkaiser FYI there was trac ticket about karma-qi roundtrip issue due to karma limiting the number of digits https://svn.boost.org/trac10/ticket/11056, I also do not think IEEE754 has anything with the string representation and AFAIK numerically stable float to string implementations produce as much digits as needed for parsers to roundtrip, see https://github.com/ulfjack/ryu.

hkaiser:
> @hkaiser FYI there was trac ticket about karma-qi roundtrip issue due to karma limiting the number of digits https://svn.boost.org/trac10/ticket/11056

Sure. I still think that there is no point in generating more digits than the internal representation can guarantee.

> I also do not think IEEE754 has anything with the string representation

I agree and I didn't mean to conflate the two things. What I meant was that if we knew that the FP numbers are IEEE 754 conforming, we could limit the precision to `15` instead of relying on `digits10`.

> and AFAIK numerically stable float to string implementations produce as much digits as needed for parsers to roundtrip, see https://github.com/ulfjack/ryu.

I need to investigate this further. I'd be happy to accept PRs that ensure numerical roundtrip stability. Just un-limiting the precision is not the correct way to go, however.

Kojoley:
> What I meant was that if we knew that the FP numbers are IEEE 754 conforming, we could limit the precision to `15` instead of relying on `digits10`.

`digits10` does not account for rounding errors, `max_digits10` should though.

> Just un-limiting the precision is not the correct way to go, however.

We clearly see here that policies that meant to allow customizing the generation are not useful because the parser overrides the value, may be move that code into a default policy?

hkaiser:
>> What I meant was that if we knew that the FP numbers are IEEE 754 conforming, we could limit the precision to 15 instead of relying on digits10.
> 
> digits10 does not account for rounding errors, max_digits10 should though.

I'd be happy to accept a PR that does this change (if it is accompanied with a thorough suite of tests for corner cases and ensuring nothing else breaks).

>> Just un-limiting the precision is not the correct way to go, however.
> 
> We clearly see here that policies that meant to allow customizing the generation are not useful because the parser overrides the value, may be move that code into a default policy?

Same here - PRs welcome.

bubnikv:
I have been bitten by this issue as well. We need to export a float into a XML file with its full precision, so after reading the float from the text file the float binary representation will not be altered. For that the std::numeric_limits<T>::max_digits10 decimal digits should be exported, possibly dropping the trailing zeros. We cannot change the numbers to a hex representation, the file format not in our control and frankly it would no more be human readable.

Boost::spirit produces one digit less than needed to guarantee a lossless round trip conversion and there is no way around it. 

Maybe karma::real_policies could be extended with a flag to decide, whether std::numeric_limits<T>::digits10 or std::numeric_limits<T>::max_digits10 is to be used to limit the number of decimal digits?


from https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10
> Conversion of a floating-point value to text and back is exact as long as at least max_digits10 were used (9 for float, 17 for double). It is guaranteed to produce the same floating-point value, even though the intermediate text representation is not exact.

We use std::stringstream, where we set 
stream << std::setprecision(std::numeric_limits<float>::max_digits10);
Using the "default" conversion policy, the stream automagically switches between scientific and fixed notation, and it produces up to the number of digits specified by setprecision, counting both the digits before and after the decimal point, thus guaranteeing the round trip conversion. I failed to achieve such behavior with karma.

Kojoley:
I still think that the PR is valid. Possibly, @hkaiser missed that the removed code hardcaps a user specified limit, the default precision of Karma was 3 digits and remains after. The change will affect only people whos custom `RealPolicy::precision` returns values bigger than this artificial hardcap. Passing test suit means that we even do not even have tests for that behavior.

hkaiser:
> Passing test suit means that we even do not even have tests for that behavior.

Fair. Let's get this in, then.

PR #589, opened 2020-04-19
Changes for Embarcadero C++ clang-based compilers, targeting Boost 1.74

eldiener:


Kojoley:
Is the old Borland compiler available for you? Maybe we simply can get rid of these workarounds if they are not already enough to make things working on those outdated compilers.

eldiener:
I only have the latest C++ Builder Community Edition 10.3. I could try to fish out from my old discs ancient editions of C++ Builder but I really do not want to bother. I could also try building with the current bcc32 compiler, but I really have little hope that it can compile current Spirit or most any other Boost library without lots of problems, and I do not want to waste any of my time trying getting bcc32 to work. It is enough work getting the Embarcadero C++ clang-based compilers to wotk with Boost libaries but at least there I have a good chance. As for getting rid of the workarounds for bcc32, that is purely up to individual maintainers. My PR should not affect any of those workarounds in any way.

include/boost/spirit/home/lex/lexer/lexertl/functor_data.hpp
```diff
@@ -68,7 +68,7 @@ namespace boost { namespace spirit { namespace lex { namespace lexertl
             {
 // some (random) versions of gcc instantiate this function even if it's not 
 // needed leading to false static asserts
-#if !defined(__GNUC__)
+#if !defined(__GNUC__) && !defined(BOOST_EMBTC)
```
Kojoley, line 71: I do not understand why this is needed, both `clang++` and `clang-cl` does not fire an assertion in
```cpp
template <typename T>
void foo(T)
{
  static_assert(false, "");
}
```
However the code itself is ill-formed and needs a rewrite...

eldiener: I was just being careful because clang on Linux and clang on Windows targeting gcc both define \_\_GNUC\_\_ and the Embarcadero C++ clang-based compilers define \_\_clang\_\_ but do not define \_\_GNUC\_\_. However I can remove this case as long as I know it does not also refer to clang. I hope you do realize that \_\_GNUC\_\_ does refer to clang also in the vast majority, unless you separately check that \_\_clang\_\_ is or is not also defined also.

eldiener: I picked up the latest 'develop' changes and merged them in my PR.

eldiener:
Can we please merge this PR ?

PR #593, opened 2020-05-04
Fix -Wdeprecated-copy

Kojoley:
Still cannot reenable -Wdeprecated-copy on CI because of Fusion, Phoenix, Proto, and Random


djowel:
Could you give me the background context what we're doing here and why we are removing these codes?

Kojoley:
Fixing https://eel.is/c++draft/depr.impldec deprecations by removing redundant code.

djowel:
Ah good! OK, looks good to me.

Issue #588, opened 2020-04-14
Some Boost.Spirit source files are not UTF-8


skelly-energid:
```
Python 3.7.5 (default, Nov 20 2019, 09:21:52) 
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: f = open("./boost/spirit/home/support/char_encoding/iso8859_1.hpp", "r")

In [2]: f.read()
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-2-571e9fb02258> in <module>()
----> 1 f.read()

/usr/lib/python3.7/codecs.py in decode(self, input, final)
    320         # decode input (taking the buffer into account)
    321         data = self.buffer + input
--> 322         (result, consumed) = self._buffer_decode(data, self.errors, final)
    323         # keep undecoded input until the next call
    324         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 10012: invalid start byte

In [3]: f = open("./boost/spirit/home/x3/support/subcontext.hpp", "r")

In [4]: f.read()
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-4-571e9fb02258> in <module>()
----> 1 f.read()

/usr/lib/python3.7/codecs.py in decode(self, input, final)
    320         # decode input (taking the buffer into account)
    321         data = self.buffer + input
--> 322         (result, consumed) = self._buffer_decode(data, self.errors, final)
    323         # keep undecoded input until the next call
    324         self.buffer = data[consumed:]

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 151: invalid continuation byte
```

djowel:
Will you do a PR?

skelly-energid:
I'm afraid not.

Kojoley:
@K-ballo could you please send a PR with your name transliterated? I am asking you about this to have done it right and to clear potential legal problems of such actions.

K-ballo:
> @K-ballo could you please send a PR with your name transliterated

@Kojoley feel free to use the ascii variant "Agustin Berge", as should already be the case in other source files.

PR #594, opened 2020-05-04
Replace invalid UTF-8 with valid UTF-8.

rubenvb:
This should fix issue #588.

I checked all of `include/` for invalid UTF-8 by running this command from the terminal when my locale was set to a UTF-8 locale:
```
grep -Raxv '.*' include/
```

### DIFF
--- a/include/boost/spirit/home/x3/support/subcontext.hpp
+++ b/include/boost/spirit/home/x3/support/subcontext.hpp
@@ -1,6 +1,6 @@
 /*=============================================================================
     Copyright (c) 2001-2014 Joel de Guzman
-    Copyright (c) 2013 Agustín Bergé
+    Copyright (c) 2013 AgustÃ­n BergÃ©
     http://spirit.sourceforge.net/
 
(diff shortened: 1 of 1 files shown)


Kojoley:
We cannot do this either because source file character mapping is implementation specific http://www.eel.is/c++draft/lex.phases#1.1
I am also afraid of altering copyright headers because I am not a lawyer.

rubenvb:
I don't quite understand why your C++ Standard quote has any relevance. They were implementation-specific bytes before, they are still now. The difference is now they are in a known, very common encoding. If you can find one supported compiler that chokes on this change (and didn't on the unchanged code), I will eat my shoe.

I can somewhat understand your copyright altering apprehension, if there was any altering going on. As it stands, this is just an encoding change. In the same sense you would be afraid to check out the Boost source code on Windows as you would alter line endings in the copyright header. If you look at the commit diff in any decent text viewer, you will see your font glyph picking can't even discern between before and after.

Maybe @K-ballo him/her/...self can weigh in on the correct spelling of his/her/... name?

PR #604, opened 2020-05-30
X3: Corner cases in alternative parser with container attribute

### LINKED ISSUE #394: X3: Attributeless parsers within container handling parsers
Update: https://svn.boost.org/trac10/ticket/6444 looks like the same problem.

X3 appends a default constructed value to containers if underlying parser of container handling parser (such as `kleene`, `plus` or `list`) does not generate anything.

```cpp
#include <boost/spirit/home/qi.hpp>
#include <vector>
#include <iostream>

template <typename F>
void test(F f)
{
    std::vector<int> v;
    char s[] = "1\n2\n3\n---\n5\n---\n---\n7\n---";
    f(std::begin(s), std::end(s), v);
    for (auto x : v) std::cout << x << '\n';
}

int main()
{
    std::cout << "qi:\n"; test([](auto f, auto l, auto& v){ qi::parse(f, l, (qi::int_ | "---") % qi::eol, v); });
    std::cout << "x3:\n"; test([](auto f, auto l, auto& v){ x3::parse(f, l, (x3::int_ | "---") % x3::eol, v); });
}
```

Output:
```
qi:
1
2
3
5
7
x3:
1
2
3
0
5
0
0
7
0
```

### LINKED ISSUE #603: Composite rules of x3::string("..."), if assigned an action, cannot be used in phrase_parse.
This is really weird and I cannot explain what's happening here. Possibly I'm stupid and not seeing something obvious, but I don't think that's the case. Here's an example of what I'm trying to do:

```cpp
    auto print_attr_type = [=](auto& ctx) {
        std::cout << typeid (x3::_attr(ctx)).name() << std::endl;
    };

    std::string s("\"Text\"");
    auto parser1 = x3::lexeme['"' >> *(x3::char_('\\') >> x3::char_('"') | x3::char_ - '"') >> '"'][print_attr_type];
    auto parser2 = x3::lexeme['"' >> *(x3::string("\\\"") | x3::char_ - '"') >> '"'][print_attr_type];
    x3::phrase_parse(s.begin(),
                     s.end(),
                     parser1,
                     x3::space);
```

Now, this compiles all fine, but if I change the parser in the parsing call to parser2, I get the weirdest compiler errors. The one where it seems to break down is

```
/usr/include/boost/spirit/home/x3/support/traits/move_to.hpp:201:24: error: no matching function for call to ‘move_to(std::remove_reference<char&>::type, std::vector<boost::variant<boost::detail::variant::over_sequence<boost::mpl::l_item<mpl_::long_<2>, std::__cxx11::basic_string<char>, boost::mpl::l_item<mpl_::long_<1>, char, boost::mpl::l_end> > > >, std::allocator<boost::variant<boost::detail::variant::over_sequence<boost::mpl::l_item<mpl_::long_<2>, std::__cxx11::basic_string<char>, boost::mpl::l_item<mpl_::long_<1>, char, boost::mpl::l_end> > > > > >&, boost::mpl::identity<boost::spirit::x3::traits::container_attribute>::type)’
  201 |         detail::move_to(std::move(src), dest
      |         ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
  202 |           , typename attribute_category<Dest>::type());
      |           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Assigning an action to `x3::string("\\\"")` alone works absolutely fine, and the definition of rule2 itself also seems to work fine, but using rule2 fails. I would be very happy if someone could help me out here, since rule2 looks nicer in code than rule1 (and should, in theory, result in a std::string attribute instead of a `deque<char,char>` one).

Kojoley:
Alternative parser were appending empty values with attributeless parsers, while it should not append anything at all.
A container parser of alternative parser of container parser with container of variant of container attribute was failing to compile.

It would be cleaner to simply store every subparser wrapped inside `alternative` because even the general way to call subparsers is via `parse_alternative`, but it will hurt parser expression tree readability.

Fixes #603
Fixes #394

### DIFF
--- a/include/boost/spirit/home/x3/operator/detail/alternative.hpp
+++ b/include/boost/spirit/home/x3/operator/detail/alternative.hpp
@@ -258,6 +258,22 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
         return false;
     }
 
+    template <typename Subject>
+    struct alternative_helper : unary_parser<Subject, alternative_helper<Subject>>
+    {
+        static bool const is_pass_through_unary = true;
+
+        using unary_parser<Subject, alternative_helper<Subject>>::unary_parser;
+
+        template <typename Iterator, typename Context
+          , typename RContext, typename Attribute>
+        bool parse(Iterator& first, Iterator const& last
+          , Context const& context, RContext& rcontext, Attribute& attr) const
+        {
+            return detail::parse_alternative(this->subject, first, last, context, rcontext, attr);
+        }
+    };
+
     template <typename Left, typename Right, typename Context, typename RContext>
     struct parse_into_container_impl<alternative<Left, Right>, Context, RContext>
     {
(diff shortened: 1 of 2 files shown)


djowel:
It's always nice that you are adding tests on changes like this. Much appreciated! Thank you for taking care of the library.

PR #608, opened 2020-06-07
Revert "Add handling with `optional<Source>` for `boost::spirit::x3::traits::…"

Kojoley:
Reverts boostorg/spirit#534

When a parser attribute type is `optional` it means that it may not produce a value, and in that case target attribute will left uninitialized. If we ever want to support this, it should be done somehow explicitly.

Also, the test does not compile, it is not run by CI because it was not add to `Jamfile`.


Kojoley:
I was thinking of putting a static assert, but there is no single good place, I probably get to it later on.

PR #616, opened 2020-07-16
Fixed signed/unsigned mismatch warnings on MinGW.

Lastique:
Use `common_type` trait and explicit casts to avoid comparing `int` and `wchar_t` constant directly, as `wchar_t` may be an unsigned type and cause warnings like this:

```
./boost/spirit/home/support/char_encoding/standard_wide.hpp: In static member function 'static bool boost::spirit::char_encoding::standard_wide::strict_ischar(int)':
./boost/spirit/home/support/char_encoding/standard_wide.hpp:86:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             return ch >= WCHAR_MIN && ch <= WCHAR_MAX;
                       ^
./boost/spirit/home/support/char_encoding/standard_wide.hpp:86:42: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
             return ch >= WCHAR_MIN && ch <= WCHAR_MAX;
                                          ^
```

### DIFF
--- a/include/boost/spirit/home/support/char_encoding/standard_wide.hpp
+++ b/include/boost/spirit/home/support/char_encoding/standard_wide.hpp
@@ -17,6 +17,7 @@
 
 #include <boost/assert.hpp>
 #include <boost/cstdint.hpp>
+#include <boost/type_traits/common_type.hpp>
 #include <boost/spirit/home/support/assert_msg.hpp>
 
 namespace boost { namespace spirit { namespace traits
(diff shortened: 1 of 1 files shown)


Kojoley:
I remember pointing @djowel about this warning when it was introduced (https://github.com/boostorg/spirit/commit/5a217a5e033d34b016fef41ee95e311bc1a24140#r35987860). It seems that the check is simply pointless and could be removed.

Btw, is not `static_cast<wchar_t>(WCHAR_MIN)` redundant?

Kojoley:
Ah, the only place where call to it is not pointless, in `toucs4` function.

Kojoley:
Nevermind, it should not be in `toucs4` in the first place.

Lastique:
> Btw, is not `static_cast<wchar_t>(WCHAR_MIN)` redundant?

It may or may not be. The point is that the type of `WCHAR_MIN` is not `wchar_t` (in this case it's `unsigned int`), so I explicitly make it `wchar_t` before the possible further conversion. Then, depending on whether `wchar_t` is signed, the conversion to `common_int` will either zero or sign extend. Whereas if you cast to `common_int` directly, the conversion will not take into account `wchar_t` signedness.

PR #620, opened 2020-08-25
Cast to correct char type to avoid assertion failure

pepsiman:
I'm getting an assertion failure in boost::spirit using boost 1.74.0:
```
boost/spirit/home/support/char_encoding/standard.hpp:59: static bool boost::spirit::char_encoding::standard::isalnum(int): Assertion `strict_ischar(ch)' failed.
```
This is being called from `boost::spirit::char_class::classify<boost::spirit::char_encoding::standard>::is<unsigned int>()`.

Comparing that function with the functions in `home/x3/support/no_case.hpp`, the wrong type is being used in the cast.

### DIFF
--- a/include/boost/spirit/home/support/char_class.hpp
+++ b/include/boost/spirit/home/support/char_class.hpp
@@ -281,7 +281,7 @@ namespace boost { namespace spirit { namespace char_class
     template <typename CharEncoding>
     struct classify
     {
-        typedef typename CharEncoding::char_type char_type;
+        typedef typename CharEncoding::classify_type char_type;
 
 #define BOOST_SPIRIT_CLASSIFY(name, isname)                                     \
         template <typename Char>                                                \
@@ -507,7 +507,7 @@ namespace boost { namespace spirit { namespace char_class
     template <typename CharEncoding>
     struct convert
     {
-        typedef typename CharEncoding::char_type char_type;
+        typedef typename CharEncoding::classify_type char_type;
 
         template <typename Char>
         static Char
--- a/test/qi/char_class.cpp
+++ b/test/qi/char_class.cpp
@@ -139,6 +139,7 @@ main()
         BOOST_TEST(test("0", xdigit));
         BOOST_TEST(test("f", xdigit));
         BOOST_TEST(!test("g", xdigit));
+        BOOST_TEST(!test("\xF1", print));
     }
 
     {


Kojoley:
Could we have a test for this?

pepsiman:
There's a commented out test in `test/qi/char_class.cpp`.

pepsiman:
Is it possible to write a portable test for `char_encoding::standard` using characters greater than 127?

Kojoley:
> Is it possible to write a portable test for char_encoding::standard using characters greater than 127?

Well, I asked the question having the same doubt and thinking that you have more expertise in the domain since you should be doing it to hit the assertion.

pepsiman:
The grammar was written by someone else and uses char_encoding::unicode everywhere, so I'm not sure why char_encoding::standard is being invoked.

If I were deliberately using char_encoding::standard, I'd expect it to allow the same inputs as the standard functions.

I suppose a test could check that no assertions fail without caring about the result of the function.

pepsiman:
Test added.

F1 is a printable character in all charsets on http://www.madore.org/~david/computers/unicode/cstab.html

sapi33:
Hi,

do you have any clue when this PR is reviewed and considered for merging?
We also faced the assertion errors in Boost 1.74.0. In our case it was due to method `isspace()` in standard.hpp, which is applied when using `boost::spirit::qi::space` in a parser. E.g. for the input 'Ä' the resulting value -67 does obviously not fulfill the condition >= 0 in `strict_ischar()`. With the PR applied the value is 195 and it seemingly works as intended.

djowel:
> do you have any clue when this PR is reviewed and considered for merging?
> We also faced the assertion errors in Boost 1.74.0. In our case it was due to method `isspace()` in standard.hpp, which is applied when using `boost::spirit::qi::space` in a parser. E.g. for the input 'Ä' the resulting value -67 does obviously not fulfill the condition >= 0 in `strict_ischar()`. With the PR applied the value is 195 and it seemingly works as intended.

From above, @Kojoley requested for a test case. Can anyone add one here so we can proceed?

djowel:
> From above, @Kojoley requested for a test case. Can anyone add one here so we can proceed?

Oh ignore that! I see a test has been added!

djowel:
Merged.

PR #621, opened 2020-08-31
Add missing include

camolezi:
This patch makes the header able to be built standalone, making possible C++ clang modules builds. @vgvassilev


camolezi:
Squashed on #626

PR #625, opened 2020-08-31
Add missing include

camolezi:
This patch makes the header able to be built standalone, making possible C++ clang modules builds. @vgvassilev


Kojoley:
Please stop this PRs spamming. I appreciate your efforts, but there is certainly a lot of files needed to be fixed and continuing like this is not an option.

camolezi:
Sorry, this is the last change that we've made. I can squash them all in one commit if it is preferable. And close the previous PRs.

djowel:
> Sorry, this is the last change that we've made. I can squash them all in one commit if it is preferable. And close the previous PRs.

Please do. And many thanks for your PR.

camolezi:
Squashed on #626

PR #627, opened 2020-10-11
Fix typos and encoding errors in copyright headers.

giomasce:


include/boost/spirit/home/x3/support/subcontext.hpp
```diff
@@ -1,6 +1,6 @@
 /*=============================================================================
     Copyright (c) 2001-2014 Joel de Guzman
-    Copyright (c) 2013 Agustin Berge
+    Copyright (c) 2013 Agustín Bergé
```
Kojoley, line 3: What is wrong here? It was intentionally ASCIIfied in 658ac180986240c75c2485df88354e83d8027b89.

giomasce: Sorry, I missed this point. I originally wrote the patch before that change was committed, and failed to notice it when I rebased it.

djowel: Can we fix this now, so we can merge this in?

giomasce: I pushed a fixed version of the commit.

djowel:
Many thanks!

Issue #511, opened 2019-06-03
attribute of x3::rule cannot be ignored since boost 1.70.0


wanghan02:
This simple code below triggers `static_assert(!has_attribute, "The rule requires an input attribute. Check your parser.")` in rule.hpp since [commit afe763b](https://github.com/boostorg/spirit/commit/afe763b61ece1a9fea5aee419920e8526e6dfcef) @Kojoley.  Let's say we have defined a rule with an attribute. We may use this rule in many places. In some places we may want to use this attribute, but in other places we may just want the boolean match result. This static_assert forces us to use the attribute every time (or attach a dummy semantic action as a workaround?...). Is it possible to remove this static_assert? Thanks!

```
#include <iostream>
#include <boost/spirit/home/x3.hpp>

namespace x3 = boost::spirit::x3;

x3::rule<struct dummy1, std::string> ruleName("name");

auto const ruleName_def = +x3::alpha;

BOOST_SPIRIT_DEFINE(ruleName);

int main() {
	std::string str("test");
	bool match = x3::parse(str.begin(), str.end(), x3::omit[ruleName]);
	std::cout << (match ? "match" : "not match") << std::endl;
    return 0;
}
```

Kojoley:
> This static_assert forces us to use the attribute every time (or attach a dummy semantic action as a workaround?...).

It is what the assertion is about. To call a rule you have to have an attribute of the type the rule was defined with, which previously was silently synthesized imposing full attribute manipulation overhead. The same situation is with semantic actions, when an outer parser does not have an attribute and you do not need it in the semantic action.

> Is it possible to remove this static_assert?

I would like to not do it. It needs to return back attribute synchronization in rules that was removed.

P.S. If rules were not strictly typed, you instead have to instantiate the rule with the all used attribute types - the different side of the coin, more flexible, but less user friendly.

wanghan02:
Let's say we have to define a very complex rule with an attribute. Do you mean we cannot use this rule in a simple match or not parser?

Kojoley:
Without synthesizing a temporary attribute - no. The assertion unrevealed a performance problem in your parser, I understand that the easiest way is to ignore it, but is not better to fix it?

djowel:
> which previously was silently synthesized imposing full attribute manipulation overhead.

[...]

PR #633, opened 2020-12-27
Test omit and raw directives behavior with rules

K-ballo:
Sanity checks to ensure `omit[]` and `raw[]` behave as expected with rules, as requested in #511.

Tested against 1.73.0 (pass) and 1.70.0 (fails due to regression). Not tested against develop/master.


djowel:
Wonderful! Looks good to me! Thanks @K-ballo !

Issue #576, opened 2020-03-05
Custom debug handler support for X3


saki7:
In Spirit.Qi, we had `qi::debug(...)` function for specifying custom debug handler. 
https://github.com/boostorg/spirit/blob/889babc3839e12ac0bba448ecd7150956d17c783/include/boost/spirit/home/qi/nonterminal/debug_handler.hpp#L75-L123

However in X3's current implementation, `simple_trace` is hard-coded:

https://github.com/boostorg/spirit/blob/889babc3839e12ac0bba448ecd7150956d17c783/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L60-L91

I'd love to see custom debug handler support in X3.

theodelrieu:
I have the same need, it's very hard to debug what's going on with all those numbers

djowel:
Would anyone care to submit a PR?

theodelrieu:
Sure, I wonder if one should be able to fully customize the handler, or if it's sufficient to add a customization point for `token_printer_debug` (in `support/traits/print_token.hpp`).

djowel:
I'd personally prefer customization points. Start with something usable but simple, then we can improve the API as needed.

theodelrieu:
I'll tackle this next week, I have to write a patch to make `char32_t` the `char_type` of the `unicode` encoding, which can have an impact on how to solve this unicode debug handler problem.

PR #634, opened 2020-12-31
Use char32_t as unicode encoding char_type

theodelrieu:
Allows to use a `symbols_parser` with unicode encoding.

### DIFF
--- a/include/boost/spirit/home/support/char_encoding/unicode.hpp
+++ b/include/boost/spirit/home/support/char_encoding/unicode.hpp
@@ -22,7 +22,7 @@ namespace boost { namespace spirit { namespace char_encoding
     ///////////////////////////////////////////////////////////////////////////
     struct unicode
     {
-        typedef ::boost::uint32_t char_type;
+        typedef char32_t char_type;
         typedef ::boost::uint32_t classify_type;
 
     ///////////////////////////////////////////////////////////////////////////
--- a/include/boost/spirit/home/support/char_set/range_functions.hpp
+++ b/include/boost/spirit/home/support/char_set/range_functions.hpp
@@ -47,14 +47,14 @@ namespace boost { namespace spirit { namespace support { namespace detail
         // another range 'other', so we can merge them
 
         typedef typename Range::value_type value_type;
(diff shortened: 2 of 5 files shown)


theodelrieu:
@djowel I did let some duplicated code in the second commit, let me know where I should put this helper ;)

Kinda fixes #576 (it only fixes the default handler).

djowel:
Nice! I've been wanting to do that for a long time! Looks good to me. I'll wait for CI to finish (hopefully it will!).

theodelrieu:
It seems that some C++03 jobs break, I couldn't reproduce on my Linux box with `cxxflags="-std=c++03"`. The logs do not mention anything related to the patch though, are those failure happening on `develop` already?

djowel:
> It seems that some C++03 jobs break, I couldn't reproduce on my Linux box with `cxxflags="-std=c++03"`. The logs do not mention anything related to the patch though, are those failure happening on `develop` already?

Does not seem to be related alright. Hmmm...

djowel:
I'll merge this now. 

The CI failures look odd. It seems the same tests are passing with the same compiler in other instances. It seems the c++03 mode are the ones that are failing. Maybe it's time to get rid of c++03 mode to lessen the burden of maintenance. In any case, the issue with test is illerevant here.

Kojoley:
No, they are not odd, the cause is `char32_t` which does not exist in C++03.

djowel:
> No, they are not odd, the cause is `char32_t` which does not exist in C++03.

Ah darned! Of course! I should have inspected the error messages more closely. Can you fix this @theodelrieu ? I'm thinking about removing support for c++03, but that will not happen anytime soon.

djowel:
> I'm thinking about removing support for c++03, but that will not happen anytime soon.

And that will involve a lot of work as well. So our best option is to fix this issue for now.

djowel:
@theodelrieu , I reverted the merge for now pending the fix for c++03: https://github.com/boostorg/spirit/pull/635

theodelrieu:
Oops, I'll fix it.

PR #636, opened 2021-01-04
Unicode/char32 t

theodelrieu:
Following up #634

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/print_attribute.hpp
+++ b/include/boost/spirit/home/x3/support/traits/print_attribute.hpp
@@ -14,6 +14,7 @@
 #include <boost/fusion/include/for_each.hpp>
 #include <boost/spirit/home/x3/support/traits/attribute_category.hpp>
 #include <boost/spirit/home/x3/support/traits/is_variant.hpp>
+#include <boost/spirit/home/support/char_encoding/unicode.hpp>
 
 namespace boost { namespace spirit { namespace x3 { namespace traits
 {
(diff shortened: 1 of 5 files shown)


theodelrieu:
@djowel Time for Windows now, unfortunately I don't have one right now, do you have an idea on why it fails with 2017 only?

djowel:
> @djowel Time for Windows now, unfortunately I don't have one right now, do you have an idea on why it fails with 2017 only?

The other fails too (2015, 2017 and 2019). Type mismatch, it seems.

theodelrieu:
>The other fails too (2015, 2017 and 2019).

Indeed... I should open my eyes before posting.

Anyway, I'll try to get it working tonight.

theodelrieu:
@djowel This is because of the `__cplusplus`, MSVC returns `199711` by default. Is there an existing macro within boost (e.g. `BOOST_HAS_CXX11` or something) that deals with this kind of things already?

djowel:
merged

include/boost/spirit/home/x3/support/traits/print_attribute.hpp
```diff
@@ -14,6 +14,7 @@
 #include <boost/fusion/include/for_each.hpp>
 #include <boost/spirit/home/x3/support/traits/attribute_category.hpp>
 #include <boost/spirit/home/x3/support/traits/is_variant.hpp>
+#include <boost/spirit/home/support/char_encoding/unicode.hpp>
```
Kojoley, line 17: I blazed through changes for the release and stuck on this line. It unconditionally includes the header which we previously only included when `BOOST_SPIRIT_X3_UNICODE` is defined.

djowel: Oh. That slipped through. We'll need to fix this for the 1.76.0 Beta candidate. @theodelrieu ?

Kojoley: I already fixed it in develop. Beta is already out, and master branch is closed, but we will have an opportunity to merge it for the release later on.

PR #644, opened 2021-02-03
Drone config

sdarwin:
Drone is a continuous integration framework similar to Travis CI. [The C++ Alliance](https://cppalliance.org/) is offering a hosted Drone server for Boost libraries. Please refer to https://github.com/CPPAlliance/drone-ci for more information and instructions.

### DIFF
--- a/.drone/drone.sh
+++ b/.drone/drone.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+
+# Copyright 2020 Rene Rivera, Sam Darwin
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.txt or copy at http://boost.org/LICENSE_1_0.txt)
+
+set -e
+export TRAVIS_BUILD_DIR=$(pwd)
+export DRONE_BUILD_DIR=$(pwd)
+export TRAVIS_BRANCH=$DRONE_BRANCH
+export VCS_COMMIT_ID=$DRONE_COMMIT
+export GIT_COMMIT=$DRONE_COMMIT
+export REPO_NAME=$DRONE_REPO
+export USER=$(whoami)
+export CC=${CC:-gcc}
+export PATH=~/.local/bin:/usr/local/bin:$PATH
+export BOOST_ROOT="$HOME/boost"
+export BOOST_BUILD_PATH="$HOME/build-boost"
+export TRAVIS_PULL_REQUEST=${DRONE_PULL_REQUEST:-false}
+export TRAVIS_REPO_SLUG=$REPO_NAME
+
+if [ "$DRONE_JOB_BUILDTYPE" == "boost" ]; then
+
+echo '==================================> INSTALL'
+
+export CACHE_NAME=$TRAVIS_OS_NAME-$TOOLSET-$STD-$JOB
+export PATH=$BOOST_ROOT:$PATH
+if [[ "$TRAVIS_COMPILER" =~ ^clang- ]]; then export STDLIB=stdlib=libc++ ; fi
+# Creating ~/user-config.jam file
+sed 's/^  //' > ~/user-config.jam << 'EOF'
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -118,6 +118,8 @@ before_install:
         -Wno-deprecated-copy      # Proto, Phoenix
       : unchecked ;
 
+    EOF
```
Kojoley, line 121: IIRC this will fail travis, please remove.

sdarwin:
@Kojoley  , you may be right.  Testing it now.  Since travis is slow, the result could take a while.   The reason for adding the EOF, is because in a typical bash script, you can't usually open a EOF section without also closing it.   Converting the same travis script to any other CI system will probably require this.   So it would be convenient to add it in travis.yml, which can act as a reference implementation.  Let's see how the test goes.

sdarwin:
Change reverted. https://travis-ci.community/t/multiline-commands-have-two-spaces-in-front-breaks-heredocs/2756

.drone/drone.sh
```diff
@@ -0,0 +1,99 @@
...
+  : unchecked ;
+
+EOF
+
+# Determining the root branch
+if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+  export BRANCH=$TRAVIS_BRANCH
+else
+  # It is a pull request. Retrieve the base branch from GitHub
+  GH_PR_API=https://api.github.com/repos/$TRAVIS_REPO_SLUG/pulls/$TRAVIS_PULL_REQUEST
```
Kojoley, line 72: These were not defined from Drone alternatives.

Kojoley:
Is Drone not running because you specified `[ci skip]` in the commit message? I login into the site and someone seems to be already enabled the repository, but build list is empty.

I see you are using boost-ci (I am not fan of that) which seems to set/override build triggers to the particular set of predefined branches what is a major PITA for me from the experience with other boost libraries/repos.

Also a question, does Drone run for forks? It is not that much of a deal for this repository as I can create a branch, but for other boost repositories I usually test my changes before opening a PR because sometime they are experimental and will not result in a PR at all.

sdarwin:
> $TRAVIS_REPO_SLUG/pulls/$TRAVIS_PULL_REQUEST

I will push a new commit here to fix it.

> Is Drone not running because you specified [ci skip] in the commit message? 

That would prevent it from running, yes.  
Also, for Drone to run the first time you must go through all the steps (not very many) in the instructions section of https://github.com/CPPAlliance/drone-ci

> which seems to set/override build triggers to the particular set of predefined branches 

triggers are now a variable. In your .drone.star, at the top of the file, add a value such as this:
```
globaltriggers={ "branch": [ "master", "develop", "drone*", "bugfix/*", "feature/*", "fix/*", "pr/*" ] }
```
Modify globaltriggers as desired. Then, in each individual job set an argument triggers=globaltriggers to use that value.

> does Drone run for forks? 

Drone is being restricted to a whitelist. Kojoley is now in the list so you may set up jobs from your forks.

djowel:
@Kojoley Tell us if we're good, and I'll merge.

Kojoley:
> Drone is being restricted to a whitelist. Kojoley is now in the list so you may set up jobs from your forks.

Much appreciated, and thanks for the PR.

sdarwin:
to avoid that EOF problem, it might be a good idea to copy the file, rather than generate it.   Put a copy of the file in a scripts/, tools/ or ci/ directory.

Here is what boostorg/beast does:

```
cp libs/beast/tools/user-config.jam ~/user-config.jam
echo "using $TOOLSET : : $COMPILER : $CXX_FLAGS ;" >> ~/user-config.jam
```

Kojoley:
Thanks for the suggestion, but I prefer having everything related to a particular CI in its configuration file for visibility.

PR #646, opened 2021-02-08
parse_rule: skip rule definition binder creation

### LINKED ISSUE #643: Why not rm the SIOF by using template metaclass specialized on rule_id?
Seth proposed a solution to a problem appearing on stackoverflow here:

  https://github.com/sehe/splitting-large-x3-parsers

Another solution was proposed here:

  https://github.com/sehe/splitting-large-x3-parsers/issues/1

which eliminated the SIOF problem with Seth's solution. 
Since Seth hasn't responded to that proposed solution, I was
hoping someone else might; hence, I'm announcing it here.
Note, this proposed solution would eliminate the need for
the rule_definition; hence, would break any existing spirit-x3
code that uses it, but that might be avoided using some sort
of #pragma.

Kojoley:
Not all of our parsers are constexpr constructible, and static local variable is not free for these.

Also, closes #643.

### DIFF
--- a/include/boost/spirit/home/x3/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/rule.hpp
@@ -195,9 +195,13 @@ namespace boost { namespace spirit { namespace x3
       , Iterator& first, Iterator const& last                                   \
       , Context const& context, BOOST_PP_CAT(rule_name, _synonym)::attribute_type& attr) \
     {                                                                           \
-        using boost::spirit::x3::unused;                                        \
-        static auto const def_ = (rule_name = BOOST_PP_CAT(rule_name, _def));   \
-        return def_.parse(first, last, context, unused, attr);                  \
+        using rule_t = BOOST_JOIN(rule_name, _synonym);                         \
+        return ::boost::spirit::x3::detail                                      \
+            ::rule_parser<typename rule_t::attribute_type, rule_t::id>          \
+            ::call_rule_definition(                                             \
+                BOOST_JOIN(rule_name, _def), rule_name.name                     \
+              , first, last, context, attr                                      \
+              , ::boost::mpl::bool_<rule_t::force_attribute>());                \
     }                                                                           \
     /***/
 #else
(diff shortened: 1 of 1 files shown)


cppljevans (changes requested):
In short, you're rm'ing the 'static auto const def' and instead inlining that def.parse.  However, not quite, AFAICT.  Shouldn't  BOOST_JOIN(rule_name, _def).rhs be the 1st arg to call_rule_definition
in order to get the RHS of the rule?

cppljevans (commented):
Why not just eliminate the static and use the rule_definition:parse directly instead of inlining it?
IOW, just this one line:

  return BOOST_JOIN(rule_name, _def).parse(first, last, context, unused, attr);

Kojoley:
> Shouldn't BOOST_JOIN(rule_name, _def).rhs be the 1st arg to call_rule_definition

No

> Why not just eliminate the static and use the rule_definition:parse directly instead of inlining it?

Because a temporary is even worse than the static local.

> IOW, just this one line:
> 
> return BOOST_JOIN(rule_name, _def).parse(first, last, context, unused, attr);

This will break annotations and debug handlers.

djowel:
@Kojoley  'Some checks were not successful", can you please check? I'd like to merge this.

cppljevans:
> 
> 
> > Shouldn't BOOST_JOIN(rule_name, _def).rhs be the 1st arg to call_rule_definition
> 
> No
> 
> > Why not just eliminate the static and use the rule_definition:parse directly instead of inlining it?
> 
> Because a temporary is even worse than the static local.
> 
> > IOW, just this one line:
> > return BOOST_JOIN(rule_name, _def).parse(first, last, context, unused, attr);
> 
> This will break annotations and debug handlers.

Ah, I see my mistake.  I had assumed, based on previous views of rule_definition use, that the
BOOST_JOIN(rule_name,_def) was an instance of rule_definition.  However, now
looking at all the examples and test (in particular, the https://github.com/Kojoley/spirit/blob/parse_rule-optimization/test/x3/rule_separate_tu_grammar.cpp), that none of the _def names are rule_definitions.
Sorry for the mistake, and of course that mistake also explains why I thought BOOST_JOIN(rule_name,_def)_.parse
would be a call to rule_definition<...>::parse.  I realize now that was a rather foolish assumption :(

Sorry for noise.

Kojoley:
> @Kojoley 'Some checks were not successful", can you please check? I'd like to merge this.

Boost.Math C++11 requirement broke some of our tests, I was waiting for CI green light to push fixes to develop branch, but Travis turnaround times these days are extreme.

PR #648, opened 2021-02-14
Replace Boost.Math with Boost.Core/std alternatives

Kojoley:
It might cost us backward compatibility with ancient compilers, exotic platforms, and yield a performance difference on MSVC, but Boost.Math is C++11 now, and Clang-cl can (or even should) be used instead of MSVC where performance matters.


djowel:
Sounds good!

PR #650, opened 2021-02-19
X3: Skip rule definition injection into context

### LINKED ISSUE #647: Any need for rule_definition?
The PR:

  https://github.com/boostorg/spirit/pull/646
  
will reduce the use of rule_definition (because it's no longer used in
the BOOST_SPIRIT_DEFINE).  The remaining uses, when used with larger
recursive grammars, will suffer the compile time explosion reported
here:

  https://sourceforge.net/p/spirit/mailman/message/37194823/
  
The compile time explosion is caused by rule_definition's storing,
when BOOST_SPIRIT_DEFINE is not used, the rule_id->RHS mapping in the
context.

  https://github.com/boostorg/spirit/blob/master/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L211
  
Since BOOST_SPIRIT_DEFINE avoids this compile time explosion, there's
no reason, AFAICT, to keep rule_definition around.  Why not remove it?

However, this removal will require a slight modification of the
default parse_rule in nonterminal/rule.hpp; however, I think that
modification would be fairly easy.  Actually, the following prototype:

```c++
//Purpose:
//  Provide a simplified analog to:
//    [rule_separate_tu]
/*
  https://github.com/boostorg/spirit/blob/master/test/x3/rule_separate_tu_grammar.hpp
  https://github.com/boostorg/spirit/blob/master/test/x3/rule_separate_tu_grammar.cpp
 */
//  which focuses solely on demonstrating how recursive Rule's are implemented.
//References:
//  [ADL] argument dependent lookup [https://en.cppreference.com/w/cpp/language/adl]
//=======================
#include <boost/preprocessor/stringize.hpp>
constexpr unsigned max_calls=10;
constexpr unsigned num_parsers=3;
#include <iostream>
namespace x3
  {
    template
    < typename Rule
    >
    constexpr bool
  defined_rule
    /**@brief
     *  Only used to avoid always firing static_assert in default parse_rule below
     */
    (
    )
    { return false;
    }
    template
    < typename Rule
    >
    bool
  parse_rule //default parse_rule
    ( Rule
    , unsigned calls
    )
    {
        std::cout<<"parse_rule(***Rule UNDEFINED***),calls="<<calls<<".\n";
        static_assert(defined_rule<Rule>(), "BOOST_SPIRIT_DEFINE undefined for this rule.");
        return false;
    }
    template
    < typename ID 
      //to enable [ADL] to work, must be defined in namespace 
      //where BOOST_SPIRIT_DEFINE is called.
    >
    struct
  rule
    {
        bool 
      parse
        ( unsigned calls=0
        )const
        {   
          std::cout<<"rule::parse(defined):calls="<<calls<<".\n";
          if(calls>max_calls) return true;
          return parse_rule(*this, calls+1);
        }
    };
  }//x3 namespace
#include <boost/preprocessor/cat.hpp>
#define BOOST_SPIRIT_DEFINE(rule_name)                          \
  bool                                                          \
parse_rule                                                      \
  ( decltype(rule_name)                                         \
  , unsigned calls                                              \
  )                                                             \
  {                                                             \
    std::cout                                                   \
      <<"parse_rule("                                           \
      <<BOOST_PP_STRINGIZE(rule_name)                           \
      <<", calls="<<calls<<").\n";                              \
    return BOOST_PP_CAT(rule_name, _def).parse(calls); \
  }                                                             \
/***/
namespace unused_attr
  {
  using namespace x3;
  //{{local struct id's 
  //  Purpose:
  //    These declarations, and their use as the rule<ID>'s
  //    in the following rule declarations, forces [ADL]
  //    to search this namespace for matching parse_rule.  These matching
  //    parse_rule's are generated by the calls to BOOST_SPIRIT_DEFINE below.
  struct id0;
  struct id1;
  struct id2;
  //}}local struct id's 
  //{{analog in [rule_separate_tu]
  //  are the declarations such as:
  //    using skipper_type = x3::rule<class skipper_r>;
  //    const skipper_type skipper;
  //  within:
  //    rule_separate_tu_grammar.hpp
  const rule<id0> rule0;
  const rule<id1> rule1;
  const rule<id2> rule2;
  //}}analog in [rule_separate_tu]
  //{{analog in [rule_separate_tu]
  //  are the declarations such as:
  //    const auto skipper_def = x3::lit('*');
  //  within:
  //    rule_separate_tu_grammar.cpp
  const auto rule0_def=rule1;
  const auto rule1_def=rule2;
  const auto rule2_def=rule0;
  //}}analog in [rule_separate_tu] 
  //{{analog in [rule_separate_tu] 
  //  is the invocations of BOOST_SPIRIT_DEFINE
  //  within:
  //    rule_separate_tu_grammar.cpp
  BOOST_SPIRIT_DEFINE(rule0)
  BOOST_SPIRIT_DEFINE(rule1)
//#define DEFINE_RULE2
#ifdef DEFINE_RULE2
  BOOST_SPIRIT_DEFINE(rule2)
#endif  
  //}}analog in [rule_separate_tu] 
  }
  int 
main()
  { unused_attr::rule0.parse(0);
    return 0;
  }
```

shows, "prototypically", how to do that.  That prototype certainly
doesn't have all of the details, such as Attribute and Iterator
template args, but that's OK because it's purpose is only to show the
"essense" of how spirit's BOOST_SPIRIT_DEFINE implements recursive
parsers.

OTOH, if the prototype is correct, then this prototype would help,
IMHO, greatly, future maintainers understand how BOOST_SPIRIT_DEFINE
and the like implement mutually recursive parsers.  I found it very
frustrating that the prototype didn't work when, instead of the
'unused_attr::id{0-2}', a simple 'unsigned ID' was used as the rule
ID.  Only after **vaguely** remembering something about how ADL works
did I mirror what spirit actually does and added the 'struct
unused_attr::id{0-2}', afterwhich, it "miraculously" worked.  I
suspect future maintainers, unless they happen to be devoted c++
language lawyers, would appreciate this prototype to show how spirit
implements mutually recursive parsers using ADL.  That's the purpose
of the code comments referencing [ADL].  The macro, DEFINE_RULE2, can
be undefined to see what happens when a rule is not
BOOST_SPIRIT_DEFINE'ed.  Quite frankly, I don't understand why the
defined_rule<Rule>() is needed within the default parse_rule.  I just
added it because, without it, the static_assertion always fired.  I
only mirrorwed what spirit did to avoid the always firing.  Maybe
someone could provide some analysis of why this defined_rule<Rule>()
is needed?

-regards,
Larry

Kojoley:
When a rule placeholder constructed and immediately consumed it cannot be used recursively, that's why the rule definition injection into a parser context can be skipped.
This optimization has a huge impact on compile times because immediate rules are commonly used to cast an attribute like `as`/`attr_cast` does in Qi.

Closes #647

### DIFF
--- a/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp
@@ -156,7 +156,7 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
         return make_unique_context<ID>(rhs, context);
     }
 
-    template <typename Attribute, typename ID>
+    template <typename Attribute, typename ID, bool skip_definition_injection = false>
     struct rule_parser
     {
         template <typename Iterator, typename Context, typename ActualAttribute>
@@ -211,7 +211,7 @@ namespace boost { namespace spirit { namespace x3 { namespace detail
             bool r = rhs.parse(
                 i
               , last
-              , make_rule_context<ID>(rhs, context, is_default_parse_rule())
+              , make_rule_context<ID>(rhs, context, std::conditional_t<skip_definition_injection, mpl::false_, is_default_parse_rule>())
               , rcontext
               , attr
             );
--- a/include/boost/spirit/home/x3/nonterminal/rule.hpp
+++ b/include/boost/spirit/home/x3/nonterminal/rule.hpp
@@ -35,10 +35,10 @@ namespace boost { namespace spirit { namespace x3
         return x3::get<ID>(context).parse(first, last, context, unused, attr);
     }
 
-    template <typename ID, typename RHS, typename Attribute, bool force_attribute_>
+    template <typename ID, typename RHS, typename Attribute, bool force_attribute_, bool skip_definition_injection = false>
     struct rule_definition : parser<rule_definition<ID, RHS, Attribute, force_attribute_>>
     {
-        typedef rule_definition<ID, RHS, Attribute, force_attribute_> this_type;
+        typedef rule_definition<ID, RHS, Attribute, force_attribute_, skip_definition_injection> this_type;
         typedef ID id;
         typedef RHS rhs_type;
         typedef rule<ID, Attribute, force_attribute_> lhs_type;
@@ -58,7 +58,7 @@ namespace boost { namespace spirit { namespace x3
         bool parse(Iterator& first, Iterator const& last
           , Context const& context, unused_type, Attribute_& attr) const
         {
-            return detail::rule_parser<attribute_type, ID>
+            return detail::rule_parser<attribute_type, ID, skip_definition_injection>
                 ::call_rule_definition(
                     rhs, name, first, last
                   , context
@@ -105,7 +105,7 @@ namespace boost { namespace spirit { namespace x3
         template <typename RHS>
         constexpr rule_definition<
             ID, typename extension::as_parser<RHS>::value_type, Attribute, force_attribute_>
-        operator=(RHS const& rhs) const
+        operator=(RHS const& rhs) const&
         {
             return { as_parser(rhs), name };
         }
@@ -113,7 +113,27 @@ namespace boost { namespace spirit { namespace x3
         template <typename RHS>
         constexpr rule_definition<
             ID, typename extension::as_parser<RHS>::value_type, Attribute, true>
-        operator%=(RHS const& rhs) const
+        operator%=(RHS const& rhs) const&
+        {
+            return { as_parser(rhs), name };
+        }
+
+        // When a rule placeholder constructed and immediately consumed it cannot be used recursively,
+        // that's why the rule definition injection into a parser context can be skipped.
+        // This optimization has a huge impact on compile times because immediate rules are commonly
+        // used to cast an attribute like `as`/`attr_cast` does in Qi.
+        template <typename RHS>
+        constexpr rule_definition<
+            ID, typename extension::as_parser<RHS>::value_type, Attribute, force_attribute_, true>
+        operator=(RHS const& rhs) const&&
+        {
+            return { as_parser(rhs), name };
+        }
+
+        template <typename RHS>
+        constexpr rule_definition<
+            ID, typename extension::as_parser<RHS>::value_type, Attribute, true, true>
+        operator%=(RHS const& rhs) const&&
         {
             return { as_parser(rhs), name };
         }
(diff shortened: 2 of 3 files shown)


cppljevans:
> 
> 
> When a rule placeholder constructed and immediately consumed it cannot be used recursively, that's why the rule definition injection into a parser context can be skipped.
> This optimization has a huge impact on compile times because immediate rules are commonly used to cast an attribute like `as`/`attr_cast` does in Qi.
> 
> Closes #647

Doesn't the statement:

  https://github.com/Kojoley/spirit/blob/x3-skip-rule-definition-injection/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L211
  
only *conditionally* "skip rule definition injection into context"?
And if that condition is not satisfied, wouldn't that still lead to
the compile-time explosion?  Or maybe, after the 1st time, recursive
calls are, somehow, coded so that NoSelfRef is always true on
recursion?  But I can't see where that happens.

Kojoley:
If you use scoped rules, it is your choice, and you pay for it. The PR fixes the cost when you are not using the feature.

cppljevans:
> 
> 
> If you use scoped rules, it is your choice, and you pay for it. The PR fixes the cost when you are not using the feature.

I'm somewhat confused by the user of several terms to mean, AFAICT,
the same thing:

* rule placeholder
* immediate rule
* scoped rule
  
I assume they all mean instances of rule_definition.  Is that right?

Now, here: 

> [attr_cast_use](https://github.com/boostorg/spirit/pull/650#issue-576658804)
  
it says:

> immediate rules are commonly used to cast an attribute 
  
which indicates an immediate rule(i.e. a rule_definition instance,
IIUC) can be used to cast an attribute.  But, as noted elsewhere, it
can be used recursively, at the cost (which is often unknown, at
first, to the end-user) of increased compile-times.  Can't the "cast
an attribute" purpose be separated from the recursive purpose?  That
seems much more desirable from the end-user perspective.  It would
also eliminate the need for the extra NoSelfRef flag:

> [rmNoSelfRef](https://github.com/Kojoley/spirit/blob/x3-skip-rule-definition-injection/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L156)

as well as the is_default_parse_rule typedef:

> [rm_def_pr](https://github.com/Kojoley/spirit/blob/x3-skip-rule-definition-injection/include/boost/spirit/home/x3/nonterminal/detail/rule.hpp#L205)
  
Of course, maybe I'm not understanding how rule_definition is used to
"cast an attribute". Could you provide an example, in particular, one
which also requires the modification of the context with the ID->rhs
mapping and thereby potentially causes the compile-time explosion?

Sorry if these questions seem simplistic, but any future maintainer of
this library would have to struggle with these same questions, in all
probability; hence, it would be nice if the answers were documented
somewhere and easily available to these hypothetical future
maintainers.

-regards,
Larry

Kojoley:
Larry, I am sorry, but requesting me to do your study every time is becoming annoying, I would prefer to spend my free time as I want to. The example is in the linked (opened by your) issue, now there are two of them, and they are fixed by this PR. The answer for 'can't users not use something like it was not suppose to be used?' is in [Hyrum's Law](https://www.hyrumslaw.com/).

PR #655, opened 2021-03-01
allow ascii symbols_parser to be matched against unicode chars

theodelrieu:
@Kojoley 

Following up on #649, I inverted the test based on your comment. Let me know what you think about it!

### DIFF
--- a/include/boost/spirit/home/x3/support/no_case.hpp
+++ b/include/boost/spirit/home/x3/support/no_case.hpp
@@ -24,8 +24,8 @@ namespace boost { namespace spirit { namespace x3
             return set.test(ch);
         }
 
-        template <typename Char>
-        int32_t operator()(Char lc, Char rc) const
+        template <typename Char, typename Char2>
+        int32_t operator()(Char lc, Char2 rc) const
         {
             return lc - rc;
         }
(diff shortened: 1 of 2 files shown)


Kojoley (commented):
My gut feeling tells me that this will open a can of worms, consider this example: `BOOST_TEST((!test(U"\U000001613", no_case[foo])))`

theodelrieu:
@Kojoley True this can lead to weird stuff like your example.

I'd like to avoid that kind of patches, but right now I don't see how to workaround my problem without writing all my grammars in UTF32.

theodelrieu:
@djowel What do you think about this one?

djowel:
I can understand @Kojoley's concern. I'm uneasy with it as well, but I do not have an easy solution ATM. Perhaps if you can start with a simple use-case, we can find something that can work better. I'm thinking something explicit, rather than implicit.

theodelrieu:
Understood, I will try to come up with a MCVE later this month.

saki7:
Closing, as this is an invalid use case involving a mix of ASCII and Unicode characters.
ASCII and Unicode differ in fundamental ways and can’t simply be interchanged through a character-type-unaware interface.

- See my comment for rationale: https://github.com/boostorg/spirit/issues/678#issuecomment-2376313379
- Other reference: <https://tonsky.me/blog/unicode/>

PR #662, opened 2021-04-02
Simplify the num list example

theidexisted:
Please merge with squash these commits.


Kojoley (commented):
I do not see the change as an improvement, could you please provide a justification of it?

theidexisted:
Reduce one line, to make the code more clear from my view.

PR #663, opened 2021-04-11
X3: Fix list parser's has_attribute

Kojoley:
List parser has an attribute only when its subject has (left operand of %).


djowel:
LGTM. Please merge as soon as CI finishes.

Issue #664, opened 2021-04-18
X3: Optional parser should fire a static assert for non-optional attributes


Kojoley:
```cpp
#include <boost/spirit/home/x3.hpp>
int main()
{
    using boost::spirit::x3::int_;
    char const* const it = "";
    int i;
    parse(it, it, -int_, i);
}
```

The problem is that after a successful parsing the attribute will be uninitialized, so accessing it will invoke UB. It is pretty much the same story as #608. Currently we cannot reliably distinguish plain attributes from optional, and just assume it could be an optional-like UDT without any checking. The `is_optional` trait needs a refinement before it can be used for implementing the static assert.

PR #665, opened 2021-04-19
X3: Optional parser is not a passthrough parser

### LINKED ISSUE #519: X3: Optional container parsing when it is a part of the sequence and the optional parser subject is a sequence too
In Boost 1.70.0, starting with `example/x3/employee.cpp` and applying the diff:
```
diff --git a/employee.cpp b/test.cpp
index 6257102..cb0fdbb 100644
--- a/employee.cpp
+++ b/test.cpp
@@ -31,7 +31,7 @@ namespace client { namespace ast
     {
         int age;
         std::string forename;
-        std::string surname;
+        boost::optional<std::string> surname;
         double salary;
     };
 
@@ -71,7 +71,7 @@ namespace client
             >> '{'
             >>  int_ >> ','
             >>  quoted_string >> ','
-            >>  quoted_string >> ','
+            >>  -quoted_string >> ','
             >>  double_
             >>  '}'
             ;
@@ -118,7 +118,6 @@ main()
 
             std::cout << "-------------------------\n";
             std::cout << "Parsing succeeded\n";
-            std::cout << "got: " << emp << std::endl;
             std::cout << "\n-------------------------\n";
         }
         else
```
results in compiler errors, including
```
/usr/include/boost/spirit/home/x3/support/traits/move_to.hpp:194:24: error: no matching function for call to ‘move_to(std::remove_reference<const char&>::type, std::__cxx11::basic_string<char>&, boost::mpl::identity<boost::spirit::x3::traits::container_attribute>::type)’
```
In contrast, the following diff does successfully compile:
```
diff --git a/employee.cpp b/test.cpp
index 6257102..79f35a0 100644
--- a/employee.cpp
+++ b/test.cpp
@@ -32,7 +32,7 @@ namespace client { namespace ast
         int age;
         std::string forename;
         std::string surname;
-        double salary;
+        boost::optional<double> salary;
     };
 
     using boost::fusion::operator<<;
@@ -72,7 +72,7 @@ namespace client
             >>  int_ >> ','
             >>  quoted_string >> ','
             >>  quoted_string >> ','
-            >>  double_
+            >>  -double_
             >>  '}'
             ;
 
@@ -118,7 +118,6 @@ main()
 
             std::cout << "-------------------------\n";
             std::cout << "Parsing succeeded\n";
-            std::cout << "got: " << emp << std::endl;
             std::cout << "\n-------------------------\n";
         }
         else
```

### Summary
The `optional` operator unary minus (`-`) works as expected when parsing a single `double_` into a `boost::optional<double>` member variable, but does not work as expected when parsing the compound parser `quoted_string` into a corresponding `boost::optional<std::string>`.

Kojoley:
Removes flat sequence parser through optional to fix:
* Non-flat `tuple<..., optional<tuple<...>>, ...>` parsing.
* Optional container parsing when it is a part of the sequence and the optional parser subject is a sequence too.

The removed test cases are invalid (see #664) variants of flat through optional. There are a valid ones (like `int_ >> -(int_ >> int_)` into `tuple<int, optional<int>, optional<int>>`) that would be broken after the change, but at the moment I do not have a popper way to make it work.

Fixes #519

### DIFF
--- a/include/boost/spirit/home/x3/operator/optional.hpp
+++ b/include/boost/spirit/home/x3/operator/optional.hpp
@@ -21,6 +21,7 @@ namespace boost { namespace spirit { namespace x3
     struct optional : proxy<Subject, optional<Subject>>
     {
         typedef proxy<Subject, optional<Subject>> base_type;
+        static bool const is_pass_through_unary = false;
         static bool const handles_container = true;
 
         constexpr optional(Subject const& subject)
--- a/test/x3/sequence.cpp
+++ b/test/x3/sequence.cpp
@@ -383,30 +383,6 @@ main()
     }
 
     // test from spirit mailing list
-    // "Optional operator causes string attribute concatenation"
-    {
-        typedef vector<char, char, int> attr_type;
-        attr_type attr;
-
-        auto node = alnum >> -('[' >> alnum >> '=' >> int_ >> ']');
-
-        BOOST_TEST(test_attr("x[y=123]", node, attr));
-        BOOST_TEST(attr == attr_type('x', 'y', 123));
-    }
-
-    // test from spirit mailing list (variation of above)
-    // "Optional operator causes string attribute concatenation"
-    {
-        typedef vector<std::string, std::string, int> attr_type;
-        attr_type attr;
-
-        auto node = +alnum >> -('[' >> +alnum >> '=' >> int_ >> ']');
-
-        BOOST_TEST(test_attr("xxx[yyy=123]", node, attr));
-        BOOST_TEST(attr == attr_type("xxx", "yyy", 123));
-    }
-
-    // test from spirit mailing list
     // "Error with container within sequence"
     {
         typedef vector<std::string> attr_type;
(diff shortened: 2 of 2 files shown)


Kojoley:
@djowel could I have your opinion on this? I probably will not be able to implement `int_ >> -(int_ >> int_)` into `tuple<int, optional<int>, optional<int>>` support in a correct way anytime soon, and potential codebrake makes me sad, even considering it is undocumented feature.

djowel:
> @djowel could I have your opinion on this? I probably will not be able to implement `int_ >> -(int_ >> int_)` into `tuple<int, optional<int>, optional<int>>` support in a correct way anytime soon, and potential codebrake makes me sad, even considering it is undocumented feature.

I was about to... Yeah, I was also wondering about the impact. But I'm thinking the bug outweighs the code break? If you agree, then go ahead.

sehe:
Hey just a head's up: the subtle change in behaviour actually surfaced a bug in some code :) See https://stackoverflow.com/questions/69896335/how-do-i-make-this-https-connection-persistent-in-beast/69908177?noredirect=1#comment136221455_69908177

Mind you - it's "behaviour change", not a "breaking change" because I think the code was relying non-atomic container attribute side-effects after backtracking. Consider it a funny anecdote, illustrating a variation on Hyrum's Law.

djowel:
> Hey just a head's up: the subtle change in behaviour actually surfaced a bug in some code :) See https://stackoverflow.com/questions/69896335/how-do-i-make-this-https-connection-persistent-in-beast/69908177?noredirect=1#comment136221455_69908177
> 
> Mind you - it's "behaviour change", not a "breaking change" because I think the code was relying non-atomic container attribute side-effects after backtracking. Consider it a funny anecdote, illustrating a variation on Hyrum's Law.

Is this something we need to undo? What is your opinion? I am on the fence.

sehe:
> Is this something we need to undo? What is your opinion? I am on the fence.

Not on my behalf! 

From my superficial glance at the issue #519 gather that the fix was legitimate? I just thought it was a fun story to share about how people will - inadvertently - depend on every observable effect of code.

In possibly related news, 

 1. I do tend to run into quite severe issues with attribute compatibility - especially getting sequences of `vector<char>` attributes to propagate into `std::string` in X3 that never used to be a problem in Qi, and 
 2. I recently ran into breakage that might also have been due to this particular change: https://stackoverflow.com/questions/45673378/boost-spirit-x3-parse-into-structs/45683933#comment135826578_45683933 - I much preferred the old style (in the answer) where rules would bind to the same attribute, instead of needing to resort to semantic actions (in the comment)

PR #669, opened 2021-04-24
X3: Fix error_handler::get_line_start

Garzet:
When the first line of input would be an empty line, `error_handler` would
fail to properly report an error due to a bug in `get_line_start` method.
Corner case check has been added to fix this.

Example:
```
std::string input = R"(
        var x int = 1 ** 2; 
)";
```
Expected:
```
In line 2:
Error! Expecting primary_expression here:
       var x int = 1 ** 2;
______________________^_
```
Produced:
```
In line 2:
Error! Expecting primary_expression here:

^_
```

### DIFF
--- a/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
+++ b/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
@@ -154,8 +154,9 @@ namespace boost { namespace spirit { namespace x3
     template <class Iterator>
     inline Iterator error_handler<Iterator>::get_line_start(Iterator first, Iterator pos) const
     {
-        Iterator latest = first;
-        for (Iterator i = first; i != pos; ++i)
+        // if first line is empty, skip it
+        Iterator latest = (*first == '\r' || *first == '\n') ? first + 1 : first;
+        for (Iterator i = latest; i != pos; ++i)
             if (*i == '\r' || *i == '\n')
                 latest = i;
         return latest;
--- a/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
+++ b/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
@@ -155,18 +155,12 @@ namespace boost { namespace spirit { namespace x3
     inline Iterator error_handler<Iterator>::get_line_start(Iterator first, Iterator pos) const
     {
         Iterator latest = first;
-        for (Iterator i = first; i != pos; ++i)
+        for (Iterator i = first; i != pos;)
             if (*i == '\r' || *i == '\n')
-                latest = i;
-
-        if (latest == first) {
-            if (*first == '\n' || *first == '\r')
-                return latest + 1;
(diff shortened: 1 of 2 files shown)


djowel:
Looks good to me. Thanks for the PR!

Kojoley:
Could you please provide a test or at least a reproducer? The solutions seems like a hack to me.

Garzet:
I've made changes to the `error_handler::get_line_start` so that is returns line start always, not only when the error occurred in the first line of the input.
I've also added a test case to `test/x3/error_handler.cpp`.

include/boost/spirit/home/x3/support/utility/error_reporting.hpp
```diff
@@ -158,7 +158,15 @@ namespace boost { namespace spirit { namespace x3
...
-        return latest;
+
+        if (latest == first) {
+            if (*first == '\n' || *first == '\r')
+                return latest + 1;
+            else
+                return latest;
+        } else {
+            return latest + 1;
+        }
```
Kojoley, line 169: I still do not understand what is going on here. Could not it be simply:
```cpp
        if (latest != pos)
            ++latest;
        return latest;
```

include/boost/spirit/home/x3/support/utility/error_reporting.hpp
```diff
@@ -199,8 +207,6 @@ namespace boost { namespace spirit { namespace x3
         err_out << error_message << std::endl;
 
         Iterator start = get_line_start(first, err_pos);
-        if (start != first)
-            ++start;
```
Kojoley, line 203: This indeed should have been inside `get_line_start`.

include/boost/spirit/home/x3/support/utility/error_reporting.hpp
Kojoley: But it is still be unsound.

include/boost/spirit/home/x3/support/utility/error_reporting.hpp
```diff
@@ -158,7 +158,15 @@ namespace boost { namespace spirit { namespace x3
         for (Iterator i = first; i != pos; ++i)
             if (*i == '\r' || *i == '\n')
                 latest = i;
```
Kojoley, line 160: I am pretty sure that the only needed fix for the function is `latest = i;` -> `latest = ++i;`.

Garzet:
Indeed -.-, sorry for waisting your time...

Kojoley:
> Indeed -.-, sorry for waisting your time...

You had spend your time dealing with the bugs in the library and tried to fix them. I appreciate it a lot.

djowel:
> > Indeed -.-, sorry for waisting your time...
> 
> You had spend your time dealing with the bugs in the library and tried to fix them. I appreciate it a lot.

Same here. And thanks for inspecting the PR, @Kojoley

PR #670, opened 2021-04-25
X3: error_handler should not skip whitespaces

### LINKED ISSUE #546: x3 error handler shows error location on the new line if an expected item is missing before the line end
If an expected item is missing before a line end, the location of the detailed error message is moved to the next line if there is a line end. Lets say you want to parse two integers separated by comma per line, then the following results in an  error message on the wrong line:

```
1
3,4
+++++++++++++++++++++++
In line 2:
Error! Expecting: delimiter ',' here:
3,4
^_
```

Instead I would expect:
```
In line 1:
Error! Expecting: delimiter ',' here:
1
_^_
```

The reason is that the error_handler<Iterator>::skip_whitespace (which is called from error_handler<Iterator>::operator()) eats the '\n' and moves the position to the next line.

An example can be found [here](https://godbolt.org/z/dWJhqF).

Kojoley:
A skipper does this already and Spirit usually does not roll back skipper work. When there is no skipper, skipping whitespaces is absolutely wrong thing to do. For example, an error points to an end of the word where an additional input is expected, but there are whitespaces after it.

The `skip_whitespace`/`skip_non_whitespace` were removed since they are not used anymore, and because they are also could trigger UB on a particular input due to unsafe `std::isspace` usage.

Fixes #546.

### DIFF
--- a/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
+++ b/include/boost/spirit/home/x3/support/utility/error_reporting.hpp
@@ -64,8 +64,6 @@ namespace boost { namespace spirit { namespace x3
         void print_file_line(std::size_t line) const;
         void print_line(Iterator line_start, Iterator last) const;
         void print_indicator(Iterator& line_start, Iterator last, char ind) const;
-        void skip_whitespace(Iterator& err_pos, Iterator last) const;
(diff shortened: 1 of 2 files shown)


djowel:
Go for it!

PR #674, opened 2021-05-06
Proper double-precision floating point hex representation of 1.0 for …

### LINKED ISSUE #673: Incorrect representation of double 1.0 for Big Endian specific tests
The representation of 1.0 for 64-bit double in IEEE 754 should be 0x3ff0 0000 0000 0000, not 0x3f80 0000 0000 0000 in the following test cases:

https://github.com/boostorg/spirit/blob/e30073e5bebb8102c73d1ad3588bf3f68cec2b64/test/qi/binary.cpp#L100
https://github.com/boostorg/spirit/blob/e30073e5bebb8102c73d1ad3588bf3f68cec2b64/test/x3/binary.cpp#L140

Looks like a simple copy-paste error, since 0x3f800... is 1.0 for single precision floats.
Also it may be hard to test this, I don't know if boost tests at https://www.boost.org/development/tests/master/developer/spirit.html includes big-endian targets. If necessary I can test on PowerPC hardware.

calvincramer:
Proper double-precision floating point hex representation of 1.0 for big-endian tests

Fixes #673

### DIFF
--- a/test/qi/binary.cpp
+++ b/test/qi/binary.cpp
@@ -97,7 +97,7 @@ int main()
             qword(0x0102030405060708LL)));
 #endif
         BOOST_TEST(binary_test("\x3f\x80\x00\x00", 4, bin_float(1.0f)));
-        BOOST_TEST(binary_test("\x3f\x80\x00\x00\x00\x00\x00\x00", 8,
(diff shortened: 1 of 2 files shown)


Kojoley:
Please double check the email address, it is either wrong or not linked to your GitHub account.

calvincramer:
I just linked the email associated with the commit in my fork

PR #691, opened 2021-06-20
Cease dependence on Foreach

Kojoley:
I have not touch examples because they are not affecting Spirit usage and testing dependencies, and modernizing them also seems like a waste of time when we encourage to start new projects with X3 instead of Qi.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ target_link_libraries(boost_spirit
     Boost::config
     Boost::core
     Boost::endian
-    Boost::foreach
     Boost::function
     Boost::function_types
     Boost::fusion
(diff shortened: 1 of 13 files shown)


djowel:
I have not looked deeper, but is there a reason why you not using the form `for(auto& x : c)` which should be a better modern substiture for `BOOST_FOREACH`?

Kojoley:
Because Qi is still supports C++03. You think we should move on even with V2 Spirit?

djowel:
Why are we moving from `BOOST_FOREACH` BTW, What did I miss?

Kojoley:
Unjustified additional dependency.

djowel:
> Unjustified additional dependency.

OK! Agreed.

Issue #668, opened 2021-04-23
double parser sometimes incorrect by one LSB


taam:
Let's start with an example:

```C++
#include <iomanip>
#include <iostream>
#include <limits>
#include <boost/spirit/include/qi.hpp>

int main() {
        using namespace boost::spirit::qi;

        std::string s = "-0.00093083996033856359";
        double x;
        if (parse(s.begin(), s.end(), double_ >> eoi, x)) {
                std::cout << s << " -> " << std::setprecision(std::numeric_limits<double>::max_digits10) << x << std::endl;
        }
}
```

Output: `-0.00093083996033856359 -> -0.00093083996033856348`

Another string/number that fails is "19403720.399719302" (`19403720.399719302 -> 19403720.399719305`), and there are many more. For all cases I've checked the binary comparison shows the mantissa being off by one. This is very bad e.g. due to (non-existing) round-trip safety.

Tested on different platforms (32 and 64 bit) with at least Boost versions 1.73.0, 1.75.0 and 1.76.0.

(Replacing `double_` with `long_double` appears to give correct results, but is a bad hack and also slow.)

djowel:
You are hitting the limits of the precision of double that is the underlying computation data type used by, double_. There were attempts to extract the last bit of precision, up to a certain point. The thing is, Spirit uses only the specified data type to provide as much speed as possible. Other parsers use higher precision data for computation, at the expense of speed. That is also why you can use long_double with ample accuracy to represent the least significant bits. In general, you need an accumulator that can provide more precision than a type T provides because floating-point computation is lossy (e.g. float -> double, double -> long_double, or perhaps some non-lossy  big-num data formats).

Kojoley:
I heard of the issue and was actually looking into a reproducer.

> In general, you need an accumulator that can provide more precision than a type T provides

Not in this case. The problem is not in accumulator, but in the target type itself. `double(93083996033856359)` is `93083996033856352.0` (`93083996033856359.0 == 93083996033856352.0`).

djowel:
> I heard of the issue and was actually looking into a reproducer.
> 
> > In general, you need an accumulator that can provide more precision than a type T provides
> 
> Not in this case. The problem is not in accumulator, but in the target type itself. `double(93083996033856359)` is `93083996033856352.0` (`93083996033856359.0 == 93083996033856352.0`).

[...]

PR #693, opened 2021-08-02
Qi double parser IEEE 754 compliance warning

taam:
This commit adds a warning regarding the issue described in #668. However, I do not know if ...
* ... a warning is appropriate,
* ... the placement is ok,
* ... the statement and wording are actually correct,
* ... maybe `float_` and `long_double` are affected as well (probably?),
* ... maybe this applies to `real_parser` in general.

It's at least a first try. I was also thinking about a longer version, but I didn't want to speculate.


Kojoley:
Does it really has something to do with IEEE 754 compliance? Unfortunately I do not have a copy of the paper, if the paper does not allow quoting at least provide a paragraph number.

taam:
I'm do not know, but I was assuming so due to your comments to that issue. I do not have the paper either. Seems I did speculate after all, sorry. That's why I was not eager to get involved into something I have no clue about.

Didn't find any hint searching the web so far, I try to look into other parser implementations if I have time.

taam:
So far I could not find anything definitive, just hints that there is a single correct conversion, e.g. https://github.com/google/double-conversion/blob/44944accfd5516a094e1e730334764e0a908aaff/double-conversion/strtod.cc very much seems to assume that there is "the correct double".

And maybe even better: https://github.com/fastfloat/fast_float/blame/main/README.md#L32-L34

djowel:
If there are tests for conformance, that is probably what we should be looking for and add to our tests. 

I offered a way to fix it: promote to a higher precision when available and only when necessary (the if-else branch). It should handle float and double. If there's no way to promote (i.e. T already has the highest precision), a proper fix would have to convert it to a quad [Quadruple precision floating-point]. The promotion from lower precision should be a lower hanging fruit. I'll have a look if I can get some free time.

saki7:
Provided improved wording.

saki7:
Thanks for your contribution! While the bug still exists, adding a warning to the documentation is a valuable first step.
Note that the bug itself is tracked on another issue, so this PR can be treated solely as a documentation fix.

PR #694, opened 2021-08-02
Update GitHub Actions CI file

sdarwin:
The Ubuntu 16.04 environment is scheduled to be removed from GitHub Actions in September 2021. Migrate those jobs to Docker containers or Ubuntu 18.04. Also, fix a problem with pip package installation on older platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,6 +309,16 @@ jobs:
       - name: Check if running in container
         if: matrix.container != ''
         run: echo "GHA_CONTAINER=${{ matrix.container }}" >> $GITHUB_ENV
+      - name: If running in container, upgrade packages
+        if: matrix.container != ''
+        run: |
+            apt-get -o Acquire::Retries=3 update && DEBIAN_FRONTEND=noninteractive apt-get -y install tzdata && apt-get -o Acquire::Retries=3 install -y sudo software-properties-common wget curl apt-transport-https make apt-file sudo unzip libssl-dev build-essential autotools-dev autoconf automake g++ libc++-helpers python ruby cpio gcc-multilib g++-multilib pkgconf python3 ccache libpython-dev
+            sudo apt-add-repository ppa:git-core/ppa
+            sudo apt-get -o Acquire::Retries=3 update && apt-get -o Acquire::Retries=3 -y install git
+            python_version=$(python3 -c 'import sys; print("{0.major}.{0.minor}".format(sys.version_info))')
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -309,6 +309,16 @@ jobs:
...
+      - name: If running in container, upgrade packages
+        if: matrix.container != ''
+        run: |
+            apt-get -o Acquire::Retries=3 update && DEBIAN_FRONTEND=noninteractive apt-get -y install tzdata && apt-get -o Acquire::Retries=3 install -y sudo software-properties-common wget curl apt-transport-https make apt-file sudo unzip libssl-dev build-essential autotools-dev autoconf automake g++ libc++-helpers python ruby cpio gcc-multilib g++-multilib pkgconf python3 ccache libpython-dev
+            sudo apt-add-repository ppa:git-core/ppa
+            sudo apt-get -o Acquire::Retries=3 update && apt-get -o Acquire::Retries=3 -y install git
+            python_version=$(python3 -c 'import sys; print("{0.major}.{0.minor}".format(sys.version_info))')
+            sudo wget https://bootstrap.pypa.io/pip/$python_version/get-pip.py
+            sudo python3 get-pip.py
+            sudo /usr/local/bin/pip install cmake
```
Kojoley, line 321: I have not had a need in this for a long while, IIRC Python 2.7, 3.4+ comes with Pip, strange to see that nowadays.

Kojoley (commented):
> The Ubuntu 16.04 environment is scheduled to be removed from GitHub Actions in September 2021. Migrate those jobs to Docker containers or Ubuntu 18.04. Also, fix a problem with pip package installation on older platforms.

Thanks, but neither of Ubuntu 16.04 or Python is used...

.github/workflows/ci.yml
sdarwin: @Kojoley , "a long while" is exactly the clue.  This script allows ubuntu 14 and ubuntu 16 to continue functioning a little while longer.   Otherwise pip is broken.   
Also, even on newer OS's, pip is a separate package, available but not automatically installed on docker containers.

Kojoley: It feels like the condition then should be `if: os == "ubuntu-16.04" or os == "ubuntu-14.04"`, but there is no 14.04 image and 16.04 will be removed soon, so there is no point in the change.

sdarwin:
> Thanks, but neither of Ubuntu 16.04 or Python is used...

Ah ha!   You're right.

I have been rolling out the update to multiple repos, automatically.  Most still use Ubuntu-16.  Closing the PR then.

PR #702, opened 2021-10-18
Fix swapped parameters to `is_substitute`

### LINKED ISSUE #679: boost::spirit::x3 problem with any ast class representing a std::string
I have two, what I presume to be legitimate, ways of defining a class in my ast that represents a string, illustrated here:
```
  struct white : std::string {};
```
or
```
  struct white {std::string text;};
```
which I use in several places for things like literals, identifiers, comments, or even white space, which I want to capture in my ast.

In some places I have to use the first form, and in other places I have to use the second form, otherwise I get compile errors with horrendous error messages:
```
  ...
  no matching function for call to ??minimal::ast::rules::rules(__gnu_cxx::__normal_iterator<minimal::ast::gap_item*,
    std::vector<minimal::ast::gap_item> >&,
    __gnu_cxx::__normal_iterator<minimal::ast::gap_item*, std::vector<minimal::ast::gap_item> >&)??
  ...
```
embedded in several pages of error output with no clue regarding the line of source code which might be at fault.

How should I interpret this type of error message?
Why do I have to use a different form/style of class in different places?

This minimal example illustrates the problem (when the #defines are [un]commented)
[rgw29_minimal.cpp.txt](https://github.com/boostorg/spirit/files/6518713/rgw29_minimal.cpp.txt)

### LINKED ISSUE #701: Compilation error in 1.77
Hi, since Boost 1.77, I am not able to compile my project that uses Spirit X3. I read in the changelog, that there were some breaking changes, but I'm not sure if that's what's affecting me. To be honest, I am kinda lost in the error message. I am not sure what it means. This is the compilation error https://gist.github.com/syyyr/3ee4bac793a0724b4976138ec35f42ed. My grammars can be seen here https://github.com/CESNET/netconf-cli/blob/master/src/grammars.hpp, the AST is in the other files next to this one, but I think that https://github.com/CESNET/netconf-cli/blob/master/src/ast_path.hpp is where the problem might be. Could you please help me solve this issue?

Kojoley:
It seems that confusion in template parameters naming lead to mistakenly swapped types to `is_substitute`. The code to X3 have been brought from Qi with the same mistake, but I do not know how to trigger it in Qi.

Fixes #701
Fixes #679

### DIFF
--- a/include/boost/spirit/home/qi/detail/alternative_function.hpp
+++ b/include/boost/spirit/home/qi/detail/alternative_function.hpp
@@ -20,22 +20,22 @@
 
 namespace boost { namespace spirit { namespace qi { namespace detail
 {
-    template <typename Variant, typename Expected>
+    template <typename Variant, typename T>
     struct find_substitute
     {
-        // Get the type from the variant that can be a substitute for Expected.
-        // If none is found, just return Expected
+        // Get the type from the Variant that can be a substitute for T.
+        // If none is found, just return T
 
         typedef Variant variant_type;
         typedef typename variant_type::types types;
         typedef typename mpl::end<types>::type end;
 
-        typedef typename mpl::find<types, Expected>::type iter_1;
+        typedef typename mpl::find<types, T>::type iter_1;
 
         typedef typename
             mpl::eval_if<
                 is_same<iter_1, end>,
-                mpl::find_if<types, traits::is_substitute<mpl::_1, Expected> >,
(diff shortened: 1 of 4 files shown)


Kojoley:
Sadly the commit https://github.com/boostorg/spirit/commit/97baf5bbe2f7c13996a6b95c8814ac458bfec419 did not put a test, and the link http://boost.2283326.n4.nabble.com/Grammar-stopped-compiling-on-more-recent-spirit-version-td3432652.html is dead

Issue #688, opened 2021-06-03
Real generator broken in 1.76


chambm:
```
#include <iostream>
#include <string>
#include <limits>
#include <algorithm>
#include <boost/spirit/include/karma.hpp>

int main()
{
    std::cout << BOOST_VERSION << std::endl;
    
    double values[] = { 0.9999, 1.9999, 2.9999, 3.9999, 4.9999, 5.9999, 6.9999, 7.9999, 8.9999, 9.9, 9.9999, 10.99, 10.9999, 11.999, 11.9999 };};
    for (auto value : values)
    {
        std::string s;
        std::back_insert_iterator<std::string> sink(s);
        boost::spirit::karma::generate(sink, boost::spirit::karma::real_generator<>(), value);
        std::cout << s << std::endl;
    }
}
```

At Coliru as of 6/3/2021, it incorrectly generates:
```
107600
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
9.0
9.9
1.0
10.99
1.0
11.999
1.0
```

At rextester it correctly generates:
```
106501
1.0
2.0
3.0
4.0
5.0
6.0
7.0
8.0
9.0
9.9
10.0
10.99
11.0
11.999
12.0
```

I'm not sure when it broke, and it's not broken for every input value, but for fairly mundane values it's very wrong. It seems to have something to do with the number of decimals places.

Kojoley:
Caused by #629. @hkaiser any ideas?

chambm:
```
                if (integer_part >= 10.)
                {
                    integer_part /= 10.;
                    ++dim;
                }
```
Looks very suspect for this bug. :) I can confirm that reverting #629 fixes the issue.

fiesh:
Oh good, there's already a ticket for this.  This hit us pretty hard in production code.

hkaiser:
> Looks very suspect for this bug. :) I can confirm that reverting #629 fixes the issue.

We'll have to find another fix for #628, then :/

RockinRoel:
My guess is that the code @chambm mentioned misses the `0 == (Policies::fmtflags::fixed & flags)` check that is used below to add the exponent.

Kojoley:
[...]

Issue #628, opened 2020-11-20
spirit::karma::generate generates 10.0e-04, but expecting 1.0e-03


markusjrmueller:
Code is:

```
std::string s;
double value = 0.00099999999999999829;
std::back_insert_iterator<std::string> sink(s);
boost::spirit::karma::generate(sink, boost::spirit::karma::real_generator<>(), value);
std::cout << s << std::endl; // outputs 10.0e-04
```

hkaiser:
The 64 bit IEEE floating point format supports only up to 15.9 digits of precision (see https://en.wikipedia.org/wiki/Double-precision_floating-point_format). The number you are trying to convert requires at least 17 digits in order to be represented. The behavior of the generator is undefined in those cases. Please see https://github.com/boostorg/spirit/issues/529 for more details.

markusjrmueller:
I am using MSCV from visual studio 2017 in 64bit. The value comes out when doing some math. They clame to be IEEE conform when not compiling with the compiler flag /fast. At the moment I use the precise-flag.
I did tests with a precision of 6 and even changed the code according to #529 but nothing changes.
I do not know what I can do to fix this issue when using karma.

hkaiser:
@markusjrmueller please see #629 for a possible fix. Please let us know if this resolves your issue.

PR #705, opened 2021-11-22
Karma: fix off by factor 10 error when rounding to integer

RockinRoel:
This fixes issue #688. There's some code that divides the integer part
by 10 and increases dim to compensate for it. This code is not
applicable to fixed notation, however, since dim is disregarded in that
case, causing the result to be off by a factor 10.

We can fix this by checking if we're not using fixed notation first,
before doing the division.

Added a few tests to check for this regression.

### DIFF
--- a/include/boost/spirit/home/karma/numeric/detail/real_utils.hpp
+++ b/include/boost/spirit/home/karma/numeric/detail/real_utils.hpp
@@ -136,7 +136,7 @@ namespace boost { namespace spirit { namespace karma
             {
                 fractional_part = floor(fractional_part - precexp);
                 integer_part += 1;    // handle rounding overflow
-                if (integer_part >= 10.)
+                if (integer_part >= 10. && 0 == (Policies::fmtflags::fixed & flags))
                 {
                     integer_part /= 10.;
                     ++dim;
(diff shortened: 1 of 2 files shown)


hkaiser (approved):
LGTM, even if this fixes only part of the issue. It overall improves things.

RockinRoel:
What part of the issue is not fixed, if I may ask? This part seemed obvious, but I don't have a good insight into everything else the algorithm is doing.

Kojoley:
> LGTM, even if this fixes only part of the issue. It overall improves things.

The release is in a week, do you think this should be included or merge the PR after the release?

> What part of the issue is not fixed, if I may ask? This part seemed obvious, but I don't have a good insight into everything else the algorithm is doing.

I explained how I understands it in https://github.com/boostorg/spirit/issues/688#issuecomment-975561527.

hkaiser:
> What part of the issue is not fixed, if I may ask? This part seemed obvious, but I don't have a good insight into everything else the algorithm is doing.

I might misremember things, but I believe that the issue is not specific to the fixed format.

RockinRoel:
Oh, you mean that it doesn't address issue #628? I indeed only looked at, and am only interested in issue #688. How numerically accurate the output is, is more important than the exact encoding to me.

PR #710, opened 2022-01-05
X3: Warn about future C++17 requirement

Kojoley:
I plan to keep the warning for two releases before starting to utilize C++17 features.

Since Boost 1.81 (scheduled to November 2022) supported compilers will be:
* Clang 4 (currently 3.6)
* GCC 7 (currently 5)
* VS 2017 (currently 2015)


djowel:
Looks good to me. It makes sense to move to c++17!

djowel (approved):

Issue #712, opened 2022-01-16
X3 3.0.10 error_handler where() is wrong


Bockeman:
The position returned by the error_handler where() points to the last successful parse point, not the beginning of the failure.  Hence the error messaging is not very meaningful and does not match the expected results in the tutorial.

Using the code from
https://www.boost.org/doc/libs/1_78_0/libs/spirit/doc/x3/html/spirit_x3/tutorials/error_handling.html

Using spirit X3 3.0.8, and in the tutorial documentation:
```
  -------------------------
  Now we have some errors
  In line 16:
  Failed: Expecting: person here:
    'I am not a person!'  <--- this should be a person
  __^_
  -------------------------
  Parsing failed
  -------------------------
```
Using spirit X3 3.0.10:
```
  -------------------------
  Now we have some errors
  In line 15:
  Failed: Expecting: person here:
    43,
  _____^_
  -------------------------
  Parsing failed
  -------------------------
```

djowel:
This is a breaking change if the behavior changed. @Kojoley, do you have an idea which commit this happened? This will break a lot of code.

Kojoley:
> @Kojoley, do you have an idea which commit this happened?

Most likely, removal of whitespace trimming (https://github.com/boostorg/spirit/pull/670).

> This will break a lot of code.

I don't think so, it is a diagnostic message for humans.

There are fundamental design choices that are conflicting: 1) sequence parser rolls back the iterator 2) error handler does not have a context to grab a skipper to skip whitespaces, but even if it had, that would be wrong because of `no_skip` existence,

I checked the idea from #704

> roll back the iterator only in alternative parser + (some?) repeat/container parsers

and it does solve the issue for `x3::error_handler`.

djowel:
OK, well, at least fix the examples, so they show how this behavior is 'properly' implemented and provide warnings of the breaking change in the comments, perhaps.

Bockeman:
The change in behaviour of the sequence parser also affects other examples which rely on `positions.position_of()` (e.g annotation.cpp).  I think this needs some careful thought.  I personally do not see any value of any position_of pointing to white space, and I would suggest that the position_of, by default, automatically advances (skips) past white space. However, I can see that there could be situations where whitespace is a genuine syntax error, so a means to override the default behaviour would be desirable.

[...]

PR #713, opened 2022-01-18
X3: Return back iterator pre-skipping for `on_success` handler

Kojoley:
Reverts the wrong part of #686. Fixes a half of #712 (`positions.position_of()`/`annotation.cpp` example part).


djowel:
Thanks @Kojoley

PR #719, opened 2022-02-18
X3: Fix inconsistent error position for `p` vs `p >> eps` or `repeat(1)[p]` in error_handler

### LINKED ISSUE #712: X3 3.0.10 error_handler where() is wrong

Kojoley:
* Do not rely on iterator rollback on fail: Preparation to relax iterator rollback policy, which is also currently violated by not rolling back skipper almost everywhere, so this should fix postskip disabling.
* Remove unneeded iterator rollback: Under the current iterator rollback policy rollbacks could be omitted in compound parser which don't advance iterator themselves because the called subparser is responsible for rollback on failure. In reality these rollbacks are undoing skipper work what renders as inconsistency in error handling position (`p` vs `p >> eps` or `repeat(1)[p]`).

Fixes #712

### DIFF
--- a/include/boost/spirit/home/x3/directive/matches.hpp
+++ b/include/boost/spirit/home/x3/directive/matches.hpp
@@ -28,8 +28,10 @@ namespace boost { namespace spirit { namespace x3
         bool parse(Iterator& first, Iterator const& last
           , Context const& context, RContext& rcontext, Attribute& attr) const
         {
+            Iterator save = first;
             bool const result = this->subject.parse(
                     first, last, context, rcontext, unused);
+            if (!result) first = save;
             traits::move_to(result, attr);
             return true;
         }
--- a/include/boost/spirit/home/x3/directive/repeat.hpp
+++ b/include/boost/spirit/home/x3/directive/repeat.hpp
@@ -76,12 +76,15 @@ namespace boost { namespace spirit { namespace x3
             }
 
             first = local_iterator;
+
+            Iterator iter = first;
             // parse some more up to the maximum specified
             for (/**/; !repeat_limit.got_max(i); ++i)
             {
                 if (!detail::parse_into_container(
-                      this->subject, first, last, context, rcontext, attr))
+                      this->subject, iter, last, context, rcontext, attr))
                     break;
+                first = iter;
             }
             return true;
         }
--- a/include/boost/spirit/home/x3/directive/repeat.hpp
+++ b/include/boost/spirit/home/x3/directive/repeat.hpp
@@ -66,16 +66,13 @@ namespace boost { namespace spirit { namespace x3
             Iterator& first, Iterator const& last
           , Context const& context, RContext& rcontext, Attribute& attr) const
         {
-            Iterator local_iterator = first;
             typename RepeatCountLimit::type i{};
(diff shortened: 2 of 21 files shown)


djowel:
This is a major move! Nice work @Kojoley. 

It would be impossible to determine the correctness by just looking at the code. So I suggest adding tests for each of the changes (if there are no such tests yet).

Kojoley:
I was planning to add more tests to have better coverage, but now I do not know when I will be able to continue the PR.
As you may know, Russia, my home country, has invaded Ukraine and currently is in an active war. I am very sad and depressed, millions of peaceful Ukrainians are suffering from inhumane decisions made by politicians which do not represent me and I had never chosen, but I still feel responsible and guilty. The only thing I could do now is to publicly say that I do not support the war with Ukraine and it should be stopped immediately.

djowel:
> I was planning to add more tests to have better coverage, but now I do not know when I will be able to continue the PR. As you may know, Russia, my home country, has invaded Ukraine and currently is in an active war. I am very sad and depressed, millions of peaceful Ukrainians are suffering from inhumane decisions made by politicians which do not represent me and I had never chosen, but I still feel responsible and guilty. The only thing I could do now is to publicly say that I do not support the war with Ukraine and it should be stopped immediately.

I have high regard and respect for you, @Kojoley. This message further affirms that. I applaud you for making a statement. Peace!

cppljevans:
I certainly want to echo Joel's statement.  I applaud your brave stance, 
and especially, the Ukrainians.

The world (except maybe just communist China and North Korea), condemn 
the invasion.

Hopefully he's not unhinged as some news reports suggest :(((

(BTW, sorry Joel for the duplicate send).

saki7:
This is a great fix. We really need to get this merged. All we need to do is resolving the conflicts and adding some unit tests. At first glance, the implementation seems ok.

saki7:
> https://github.com/boostorg/spirit/issues/833#issuecomment-3333749234
> 
> It's been quite a while and I don't remember much, it's most likely the case https://github.com/boostorg/spirit/pull/719 was supposed to fix. I remember I almost finished the patch but haven't updated the PR. I can try to find and upload my latest version of that PR, but I'm not going to rebase and resolve merge conflicts that emerged from the code churn that has happened since then, sorry.

@Kojoley Could you push the recent version? If you don't have time then feel free to push it as-is. I'm not entirely sure if it can be merged into feature-frozen X3, but your work is still important and I might be able to adopt it to the X4 codebase.

PR #727, opened 2022-06-27
X3: fix missing template argument skip_definition_injection for rule_defi…

wanghan02:
template argument `skip_definition_injection` is missing when defining `struct rule_definition` and trying to refer to itself as `Derived` in `parser<Derived>`.


wanghan02:
Hi @Kojoley , I'm not sure if you are still maintaining spirit. Do you know why did my change in "[include/boost/spirit/home/x3/nonterminal/rule.hpp]" break checks for qi and karma?

Kojoley:
It's unrelated to your change. Thanks for catching the bug.

wanghan02:
> It's unrelated to your change. Thanks for catching the bug.

Thanks!

PR #732, opened 2022-08-30
Fix typo in #include composite.hpp

georgthegreat:


include/boost/spirit/home/classic/utility/scoped_lock.hpp
```diff
@@ -12,7 +12,7 @@
 #include <boost/spirit/home/classic/namespace.hpp>
 #include <boost/thread/lock_types.hpp>
 #if !defined(BOOST_SPIRIT_COMPOSITE_HPP)
```
Kojoley, line 14: @djowel do you have any idea why there is that `#if`?  It predates migration from CVS, as well as the include guard in the included file.

georgthegreat:
NB: we have found this typo by using the automated dependency parser for C++.
Not sure if this #include ever works

Kojoley:
> Not sure if this #include ever works

It could be hit by directly including the header which is explicitly allowed (https://www.boost.org/doc/libs/1_79_0/libs/spirit/classic/doc/includes.html), though it seems no one did that or reported the issue.

include/boost/spirit/home/classic/utility/scoped_lock.hpp
djowel: I have no idea, or perhaps I already forgot.

Kojoley: I am going to make the include unconditional then, cannot imagine why it would be wrong.

Kojoley:
Thanks for the catch!

PR #733, opened 2022-09-08
Avoid some of the deprecated copy warnings

apolukhin:
The patch fixes only some of the warnings. Remaining warnings could be located via `./b2 libs/spirit/test/ cxxstd=17 "cxxflags=-Werror=deprecated-copy" toolset=clang-15 -j4`

### DIFF
--- a/include/boost/spirit/home/support/unused.hpp
+++ b/include/boost/spirit/home/support/unused.hpp
@@ -27,6 +27,8 @@ namespace boost { namespace spirit
     struct unused_type
     {
         BOOST_DEFAULTED_FUNCTION(unused_type(), {})
+        BOOST_DEFAULTED_FUNCTION(unused_type(const unused_type& ), {})
+        BOOST_DEFAULTED_FUNCTION(unused_type& operator=(const unused_type& ), { return *this; })
 
         template <typename T>
         unused_type(T const&)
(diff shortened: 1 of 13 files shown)


Kojoley:
Thanks, I was fixing these earlier this year but got sidetracked by fixing it in the dependencies first, there still warnings coming from Fusion and I have a patch for it I need to file upstream. I don't like that we change type triviality to suppress a useless warning, so I had been replacing fake-assign operators with just warning suppression.

include/boost/spirit/home/support/unused.hpp
```diff
@@ -27,6 +27,8 @@ namespace boost { namespace spirit
     struct unused_type
     {
         BOOST_DEFAULTED_FUNCTION(unused_type(), {})
+        BOOST_DEFAULTED_FUNCTION(unused_type(const unused_type& ), {})
+        BOOST_DEFAULTED_FUNCTION(unused_type& operator=(const unused_type& ), { return *this; })
```
Kojoley, line 31: I don't think these are needed...

apolukhin: Reverted

Kojoley:
@apolukhin I'm waiting on https://github.com/boostorg/phoenix/pull/108 to be merged.

PR #736, opened 2022-09-16
Karma: real generator off by a magnitude due to log10 rounding up

### LINKED ISSUE #529: karma::real_policies<double>::fraction_part() formats incorrectly for precision >= 16 for fractions whose digits form an integer >= 999999999999998

### LINKED ISSUE #735: Small bug in karma real_generator output.
If I try to output number 0.099999999999999770 (double) with 16 digits precision, boost karma generates wrong output of "0.9999999999999999" which is 10 times larger. As far as I can tell this is an issue of log10 rounding, which for this number produces 15.0 exactly instead of 14.999.... in real_policies.hpp : fraction_part. This causes karma to not output a leading zero.

I'm not exactly sure, but I think this issue can be resolved by replacing floor(log10(n)) + 1 with ceil(log10(n)). Also there is probably other places in real generator where such issues can arise.

Kojoley:
Closes #735
Closes #529

### DIFF
--- a/include/boost/spirit/home/karma/numeric/real_policies.hpp
+++ b/include/boost/spirit/home/karma/numeric/real_policies.hpp
@@ -258,7 +258,7 @@ namespace boost { namespace spirit { namespace karma
             // but it's spelled out to avoid inter-modular dependencies.
 
             typename remove_const<T>::type digits = 
-                (traits::test_zero(n) ? 0 : floor(log10(n))) + 1;
+                (traits::test_zero(n) ? 1 : ceil(log10(n + T(1.))));
             bool r = true;
             for (/**/; r && digits < precision_; digits = digits + 1)
                 r = char_inserter<>::call(sink, '0');
--- a/test/karma/real3.cpp
+++ b/test/karma/real3.cpp
@@ -36,6 +36,11 @@ namespace boost { namespace spirit { namespace traits
 }}}
 #endif
 
+struct double_prec16_policy : boost::spirit::karma::real_policies<double>
+{
+    static unsigned int precision(double) { return 16; }
+};
+
 ///////////////////////////////////////////////////////////////////////////////
 int main()
 {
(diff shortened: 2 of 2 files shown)


Kojoley:
I'll merge this on weekend if there is no objection. Better to try out the fix than to do nothing.

hkaiser (approved):
LGTM, thanks!

friedrichatgc (commented):
This PR leads now to wrong outputs of other values:
for example 3.1 is output as the string "**3.01**000000000000001"

This still only happens if the precision is >=16.
(but a precision of 16 is needed for example to get bit-identical values for float-string-float round-trips; so restricting to a precision of 15, as also discussed, is no option for me)

Issue #610, opened 2020-06-08
X3: Alternative parser attribute in presence of attribute-less branches


Kojoley:
In #607 I rewrote the alternative parser attribute deducer to not use MPL, and while it is equivalent to the original, I forgot that the original was very questionable.

It:
1) Does not remove duplicates, so for `int_ | int_` the attribute will be `variant<int, int>` while it should be `int`
2) It removes `unused_type`, so for `int_ | eps` the attribute will be `int`

```cpp
template <typename Parser, typename Context = boost::spirit::x3::unused_type>
void print_parser_attribute_type(Parser const&)
{
    using attr_type = typename boost::spirit::x3::traits::attribute_of<Parser, Context>::type;
    std::cout << boost::core::demangle(typeid(attr_type).name()) << '\n';
}

int main()
{
    using namespace boost::spirit::x3;
    print_parser_attribute_type(int_ | int_);
    print_parser_attribute_type(int_ | eps);
}
```

While deduplication is a clear and simple issue, the `unused_type` part is tricky, and it also goes against the documentation https://www.boost.org/doc/libs/develop/libs/spirit/doc/x3/html/spirit_x3/quick_reference/compound_attribute_rules.html#spirit_x3.quick_reference.compound_attribute_rules.compound_parser_attribute_types:

```
a: A, b: B --> (a | b): variant<A, B>
a: A, b: Unused --> (a | b): optional<A>
a: A, b: B, c: Unused --> (a | b | c): optional<variant<A, B> >
a: Unused, b: B --> (a | b): optional<B>
a: Unused, b: Unused --> (a | b): Unused
a: A, b: A --> (a | b): A
```

However, I propose to not turn `int_ | bool_ | eps` into `optional<variant<int, bool>>`, instead leave `unused_type` and make it the first type in the variant, so for `int_ | bool_ | eps` the attribute is `variant<unused_type, int, bool>`. Making the first type `unused_type` means for variant that `unused_type` is constructed in default constructor, this is what we needed when the branch does not assign the value. Probably we need a trait when someone wants `std::monostate` instead of `unused_type`.
For the case when alternative parser has an attribute of `variant<unused_type, T>` convert it to `optional<T>`, nothing new here, the same as in the documentation.

saki7:
Godbolt for the current codebase: <https://godbolt.org/z/v79P53jds>

---

> 1. Does not remove duplicates, so for `int_ | int_` the attribute will be `variant<int, int>` while it should be `int`

[...]

PR #740, opened 2022-10-30
X3: Rework attribute for alternative parser

eido79:
Implement the following behavior, as described in #610 
* Remove duplicates
* Simplify `variant<X>` to `X`
* Make `unused_type` the first type of the variant
* Transform `variant<unused_type, T>` to `optional<T>`


Kojoley (commented):
Thanks for the PR. A few things need to be resolved:
1) I haven't made deep inspection of what could go wrong with doing `variant<unused_type, ...>`, have you? At least the question about how crude is `unused_type` in visitor will be, since it has implicit constructor from everything.
2) A trait to configure monostate type is needed.
3) No need to use boost type traits which are available in standard library.
4) Put the tests in alternative.cpp/sequence.cpp please.

eido79:
Hi @Kojoley,

> I haven't made deep inspection of what could go wrong with doing variant<unused_type, ...>, have you? At least the question about how crude is unused_type in visitor will be, since it has implicit constructor from everything.

Not really, admittedly. The fact that unused_type is constructible from everything might indeed create problems. I'll do some more analysis and try to come up with some meaningful use cases.

In case this turn out to be ugly, it is anyway easy to make the behavior of `attribute_of_alternative` match the currently documented rules, i.e. use `optional<variant<...>>` when `unused_type` is involved.

> A trait to configure monostate type is needed.

Agreed. How do you think that should look like?

> No need to use boost type traits which are available in standard library.
> Put the tests in alternative.cpp/sequence.cpp please.

OK.

Kojoley:
> > I haven't made deep inspection of what could go wrong with doing variant<unused_type, ...>, have you? At least the question about how crude is unused_type in visitor will be, since it has implicit constructor from everything.
> 
> Not really, admittedly. The fact that unused_type is constructible from everything might indeed create problems. I'll do some more analysis and try to come up with some meaningful use cases.
> 
> In case this turn out to be ugly, it is anyway easy to make the behavior of `attribute_of_alternative` match the currently documented rules, i.e. use `optional<variant<...>>` when `unused_type` is involved.

I am afraid `optional<variant<...>>` might actually be not supported currently by alternative parser/is_subsitute, I couldn't find tests for it either.

> > A trait to configure monostate type is needed.
> 
> Agreed. How do you think that should look like?

Something like https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/x3/support/traits/optional_traits.hpp
```cpp
template <typename Variant> struct variant_monostate { using type = unused_type; };
```

eido79:
It's indeed ugly. Since anything is convertible to `unused_type`, things like this compile just fine
```
char input[] = "Whoops";
boost::variant<x3::unused_type, int> v;
auto first = std::begin(input);
auto last = std::end(input) - 1;
bool success = x3::parse(first, last, x3::int_ | +x3::char_, v);
assert(success && first == last); // OK!
```
hiding potential mistakes that should be caught at compile time.

At this point, I'm thinking that using `boost::blank` or another empty type would be a better choice.
Alternatively, it could be investigated what would it take to make `optional<variant<...>>` work properly.

Thoughts?

saki7:
@Kojoley Is this issue still valid on current codebase?

saki7:
Closing this for several reasons:

1. The PR is stale, and no improvements were made since 2022.
2. See my comments on https://github.com/boostorg/spirit/issues/610#issuecomment-3014781535

Issue #698, opened 2021-08-30
Is there a `x3::position_tagged` without inheritance?


denzor200:
Hi everyone! Is it possible to add the `x3::position_tagged` attribute to a custom structure without applying inheritance? If it's not possible, it question should resolve, because in the future BOOST PFR library maybe will be used on the Spirit side (The corresponding PR has already been created in the neighboring Fusion repository)

Recall that the pfr library is not able to work with inherited structures, the presence of inheritance for this library is not acceptable.

If this issue has not been resolved yet, I suggest solving it using composition instead of inheritance:
```
struct employee
{
	int age;
	person who;
	double salary;
        x3::position_tagged __position_tagged;
};

/// The library is aware of all the fields as they are listed in the BOOST_FUSION_ADAPT_STRUCT macro.
/// Among these fields, a search is performed for the field whose type corresponds to x3::position_tagged, the name of the field is not important.
/// Now this construction looks strange, but in the future the fields will be automatically detected using the PFR. The BOOST_FUSION_ADAPT_STRUCT macro will not be required
BOOST_FUSION_ADAPT_STRUCT(client::ast::employee,
    age, who, salary, __position_tagged
)
```

saki7:
Closing as per https://github.com/boostorg/spirit/pull/742#issuecomment-2865859762

PR #742, opened 2022-11-23
X3: position_tagged without inheritance

denzor200:
This PR resolves https://github.com/boostorg/spirit/issues/698.
It needs to make Spirit more friendly with PFR limitations.

I did this feature only and only for sequence parser(I can't find any reason to make it for another parsers).
Also, this feature won't support size one sequences.


saki7:
I'm afraid that this approach is not robust and thus potentially introducing another layer of complexity on X3's implementation details.

I am also one of the users who are disturbed by the intrusive nature of `position_tagged`. I've even created my own utility for tracking positions non-intrusively. That kind of workaround can be done by just defining your own error handler, since all rules will eventually be passed onto the handler, which essentially provides you with an opportunity to *track* positions for any rules.

Your approach seems to require injecting `position_tagged` on the attributes. Even if this approach is seemingly working on your specific code, I'd rather think the solution is way too complicated than the nature of the issue. There are more fundamental issues regarding the composition and subsumption of variant substitutes, which can be affected by exposing `position_tagged` as an alternative attribute.

PR #756, opened 2023-03-08
line_pos_iterator on-top of bidirectional-iterators

tobias-loew:
I came over line_pos_iterator and really liked the example from https://stackoverflow.com/questions/19612657/boostspirit-access-position-iterator-from-semantic-actions. The only problem, finding the line-start is linear in the input, which is a burden for large inputs. But when the iterator is bidirectional, finding the line-start can be implemented linear to the average line-length.


djowel:
Nice!

djowel:
Wouldn't it be nice to have this incorporated into the original line_pos_iterator and just switch to the code depending on iterator type? That's kinda what I was expecting.

tobias-loew:
I'm on it.

PR #765, opened 2023-07-20
prefer auto to Rng::(const_)?iterator for generic String

wanghan02:
1. c++20 ranges library does not require a range to have member type/alias iterator/const_iterator.
2. boost range library uses `typename boost::range_iterator<Rng>::type` instead of `Rng::iterator` because a customized Rng may not have member type/alias iterator/const_iterator. e.g. `std::pair<It, It>`.

Therefore for generic String type, `auto` is better than `Rng::(const_)?iterator`.

### DIFF
--- a/include/boost/spirit/home/x3/support/traits/string_traits.hpp
+++ b/include/boost/spirit/home/x3/support/traits/string_traits.hpp
@@ -133,21 +133,19 @@ namespace boost { namespace spirit { namespace x3 { namespace traits
 
     // Implementation for containers (includes basic_string).
     template <typename T, typename Str>
-    inline typename Str::const_iterator get_string_begin(Str const& str)
+    inline auto get_string_begin(Str const& str) -> decltype(str.begin())
     { return str.begin(); }
 
     template <typename T, typename Str>
-    inline typename Str::iterator
-    get_string_begin(Str& str)
+    inline auto get_string_begin(Str& str) -> decltype(str.begin())
     { return str.begin(); }
 
     template <typename T, typename Str>
-    inline typename Str::const_iterator get_string_end(Str const& str)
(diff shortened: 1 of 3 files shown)


include/boost/spirit/home/x3/support/traits/string_traits.hpp
wanghan02, line 1: trailing return type `-> decltype(boost::(begin|end)(str))` is a more accurate SFINAE condition than `typename Str::(const_)?iterator`

wanghan02:
Somehow CI with older MSVC versions failed on karma but the changes in this commit is for x3 only.

```
.\boost/spirit/home/classic/core/primitives/primitives.hpp(145): error C2338: N < 3
libs\spirit\classic\test\char_strings_test_fail.cpp(18): note: see reference to function template instantiation 'boost::spirit::classic::chlit<char> boost::spirit::classic::ch_p<char,7>(const CharT (&)[7])' being compiled
        with
        [
            CharT=char
        ]
```

wanghan02:
The changes in this commit are all in x3 folder with std14 at least. Somehow CI fails in some std03/11 karma/qi builds.

saki7:
I'm afraid that this is not the ideal solution for upgrading the codebase. For instance, C++20 ranges have `std::ranges::iterator_t` which is advised to be used when extracting such traits.

Also, a quick search on `x3/` directory shows that we have more non-trivial issues regarding the legacy/modern mismatch of range handlings, such as: https://github.com/boostorg/spirit/blob/ba53544ac90778a7f2774c21fb552572bf7b37bd/include/boost/spirit/home/x3/support/traits/is_range.hpp#L16-L24

If we are to do the modernization on range manipulations, we should apply the fix comprehensively. This PR might surely solve some issues, but a lot more work should be done (sadly).

Feel free to reopen if you have a better implementation.

PR #766, opened 2023-09-02
Switch to `boost::core::invoke_swap`

Lastique:
`boost::swap` is deprecated and will be removed. Use `boost::core::invoke_swap` as a replacement, where needed.

Removed `using boost::swap` in `common.hpp`, since this is not the intended usage of Boost.Swap, and doing so did not add anything to the existing code (which already did `using std::swap`).

Corrected documentation re. compatibility with `boost::swap`/`boost::core::invoke_swap`. It does not (and never did) use member `swap()` functions, but simply forwards to either `std::swap` or the user-defined free function overload (which, in turn, may forward to the member `swap()`).


Lastique:
The errors caused by undefined `std::ios_base` should be fixed by #767.

PR #770, opened 2023-09-22
Update syntax examples to parser operator docs

### LINKED ISSUE #769: X3 docs for operators is missing syntactical expressions
The page describing [Parser Operators](https://www.boost.org/doc/libs/1_83_0/libs/spirit/doc/x3/html/spirit_x3/quick_reference/operator.html) is missing the syntactical expressions in the leftmost column, e.g. it just says "Expectation Operator" instead of showing `a > b`.

Krzmbrzl:
Fixes #769


Krzmbrzl:
The failing CI seems to be due to some code using deprecated constructs? Since this PR doesn't change any code, I guess I can just ignore the CI?

saki7:
Huge thanks for this PR! This issue was one of the very first challenges I ran into when I was just getting started with Spirit. Improvements like this are exactly what we need to help make the library more approachable.

PR #783, opened 2024-02-07
Fix missing includes for karma/binary

iv-m:
Previosuly, scoped_enum was included into karma's binary.hpp through boost/endian/detail/order.hpp. Since Boost 1.84.0 that file stopped using scoped enum and switched to use plain enum class instead; so it needs to be added here.


iv-m:
I discovered this issue when trying to build some code with boost 1.84.0. With boost 1.83.0 it compiled just fine, but with 1.84.0 it fails with

```
/usr/include/boost/spirit/home/karma/binary/binary.hpp:197:19: error: 'BOOST_SCOPED_ENUM' has not been declared
  197 |         template <BOOST_SCOPED_ENUM(boost::endian::order) bits>
      |                   ^~~~~~~~~~~~~~~~~
```

... and  a few screens of various errors.

PR #785, opened 2024-02-23
X3: Constrain `operator>`

### LINKED ISSUE #782: MSVC (VS 2023) fails to compile example/x3/annotation.cpp with /std:c++20 or /std:c++latest option
MSVC fails to compile [annotation example ](https://www.boost.org/doc/libs/1_83_0/libs/spirit/example/x3/annotation.cpp) if  /std:c++20 or /std:c++latest is set. Everything is fine with /std:c++17.

See [compiler explorer](https://godbolt.org/z/hGs9eEr54) to get the bug reproduction

Unfortunately the error message ( can be seen in the provided link ) is pretty cryptic so I'm not sure whether it's a MSVC or boost::spirit problem.

Kojoley:
It might be instantiated for user types inherited from classes in our namespace.

Fixes #782

### DIFF
--- a/include/boost/spirit/home/x3/operator/sequence.hpp
+++ b/include/boost/spirit/home/x3/operator/sequence.hpp
@@ -59,6 +59,7 @@ namespace boost { namespace spirit { namespace x3
 
     template <typename Left, typename Right>
     constexpr auto operator>(Left const& left, Right const& right)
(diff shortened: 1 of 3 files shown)


djowel:
Looks good, @Kojoley.

Issue #792, opened 2024-09-19
karma::real_policies<double>::fraction_part() formats incorrectly by order of magnitude for numbers like 1.09e-6, emits 1.9e-06(!)


sigbjorn:
# Overview

As detected reported on  Shyft open source, https://gitlab.com/shyft-os/shyft/-/issues/1216,

**Wrong outputs** are generated for numbers of the form **1.09e-6**,  instead of generating expected '1.09e-06',
it outputs **1.9e-06**
-lacking trailing zeros before the 9, so an entirely different number, not resembling the input (like rounding errors etc).

This is due to an edge case not detected in the test-setup, where the passed n to the fraction routine is of the form:

9999.00000005 or similar.

## How to reproduce

Adding the following test snippet to the test-suite, karma/real3.cpp:

```
    // test for off by a magnitude due to log10 policy fraction for patterns of 999.00 rounding up
    {
        BOOST_TEST(test("1.09e-06",
			double_,
            1.09e-6
            ));
    }
```
Failes the test, with output
```
in test: result is false
in test: generated "1.9e-06"
karma/real3.cpp(202): test 'test("1.09e-06", double_, 1.09e-6 )' failed in function 'int main()'
1 error detected.

EXIT STATUS: 1
```

### Minimal
The minimal fix (in terms bytes), is to use floor(n) in the digit expression to get rid of trailing digits that promotes to computing one to much digits.

[real_policies.hpp](https://github.com/boostorg/spirit/blob/b44bcac5259877633b1a6c5f662898562a66f9c2/include/boost/spirit/home/karma/numeric/real_policies.hpp#L261)
```
            typename remove_const<T>::type digits = 
	      (traits::test_zero(n) ? 1 : ceil(log10(floor(n) + T(1.))));
```
The idea is to remove the fraction part of the fraction passed (so 9.0000.....5 becomes 9.0 and the log10..ceil provides 1 in stead of 2, and thus one 0 is injected before the 9 digit.

### A bit more verbose solution

Other approach is based on example provided by other issues around the same function, where the computation of digits needed to format an int is factored out.

A mod of that example, in-lined, and written so that it preserves the real_concept used in the tests, so that the test-suite passes,
could look like this: 

```
    template <typename OutputIterator>
    bool fraction_part(OutputIterator& sink, T n, unsigned int precision_, unsigned int precision) const
    {
      using namespace std;
      unsigned int digits=1;
      if(! boost::spirit::traits::test_zero(n)) {
[...]

PR #793, opened 2024-09-19
Proposal for solving fraction_part causing 1.09e-6 generated as 1.9e-6 issue  #792

sigbjorn:
Solves issue, 

based on using int based compute of number of digits in a positive real-concept type,
as provided in Shyft workaround by 
 Paul Christoph Bätzing-Rosenvinge: https://gitlab.com/paul.christoph.batzing-rosenvinge
 
Ref to #792


djowel:
@hkaiser Tell me what you think?

hkaiser:
> @hkaiser Tell me what you think?

I'm fine with this if all the tests pass.

saki7:
I'm not entirely convinced this is the best approach. There's likely a better way than dividing by 10 in a loop to analyze decimal digits. That said, a more optimized implementation would require a deep understanding of IEEE floating-point representation. For now, I think this is an acceptable solution, as it addresses a real issue effectively.

Thanks for the PR!

PR #803, opened 2025-05-09
Turn off -Werror in order to deal with unfixable upstream warnings

saki7:
Make -Werror off, as it can't handle upstream warnings properly.
Specifically, container and math is emitting warnings that makes spirit unmaintainable since all tests fail.

Also includes some QoL fixes on the CI configuration.


saki7:
Before/after

image

image

PR #806, opened 2025-06-08
Update sexpr_parser.hpp

rcoca:
updated rule for list - to have it captured in utree::list_type - and thus preserve the actual structure of the parsing tree


saki7:
I tested the example, and it turns out the code compiles with or without the fix.

`utree::list_type` has a conversion constructor that allows transparent construction from `utree`, so there's no compelling reason to expose the *internal* list type in the example. While your version looks more strict in terms of parser attributes, I'm not convinced that's a sufficient reason to change a 15-year-old example.

Since I haven't received any further information since my last comment, I'm closing this for now. If you still believe the change is necessary, please explain your reasoning.

rcoca:
Sorry, I saw your message late.
I trust you judgement here, the parser runs just fine and it does syntax check.
It just builds a list - instead of an AST. But that's probably not example's aim.

Issue #346, opened 2017-12-31
X3: MPL usage


Kojoley:
Currently X3 uses MPL which is still not updated for C++11. It is slow, even including it adds a big time overhead (see benchmarks on [hana docs](http://www.boost.org/doc/libs/1_65_0/libs/hana/doc/html/index.html)), but actually it is used intensively only in a few places.
What if we rewrite them without MPL?

Example:
https://github.com/boostorg/spirit/blob/49587ff7a04d0301d798d080e541f45debbb737f/include/boost/spirit/home/x3/operator/detail/sequence.hpp#L237-L294

This could be easily done without any meta-library (and twice shorter):
```cpp
// a helper class, may be reused in multiple places
template <typename... T>
struct type_pack
{
    template <template <typename...> class U>
    using transfer = U<T...>;

    template <typename... U>
    using prepend = type_pack<U..., T...>;
};

// this class also can be shared by alternative parser
template <typename T, typename Attribute>
struct attribute_prepender { using type = typename T::template prepend<Attribute>; };
// filter unused attribute
template <typename T>
struct attribute_prepender<T, unused_type> { using type = T; };
// a shortcut
template <typename T, typename P, typename C>
using prepend_parser_attribute = typename attribute_prepender<T, traits::attribute_of_t<P, C>>::type;

template <typename C, typename L, typename... T>
struct get_sequence_types;
// place underlying parsers of sequence_parser to the queue
template <typename C, typename T, typename L, typename R, typename... U>
struct get_sequence_types<C, T, sequence_parser<L, R>, U...> : get_sequence_types<C, T, R, L, U...> {};
// prepend an attribute of parser to type pack
template <typename C, typename T, typename H, typename... U>
struct get_sequence_types<C, T, H, U...> : get_sequence_types<C, prepend_parser_attribute<T, H, C>, U...> {};
// we are done, return the pack
template <typename C, typename T>
struct get_sequence_types<C, T> : T { };

template <typename L, typename R, typename C>
struct attribute_of_sequence
{
    using type = typename get_sequence_types<C, type_pack<>, R, L>::template transfer<fusion::deque>;
};
```

Proof of concept http://coliru.stacked-crooked.com/a/5368077425eec7e6

<details>
<summary>Proof of concept code</summary>

```cpp
#include <type_traits>
#include <tuple>
#include <typeindex>
#include <iostream>
#include <boost/core/demangle.hpp>

template <typename... T>
struct type_pack
{
    template <template <typename...> class U>
[...]

PR #812, opened 2025-09-06
Use concepts for attribute category overload resolution

### LINKED ISSUE #270: Add support for std::optional
It would be nice if Boost.Spirit recognized `std::optional` in C++17 the way it does `boost::optional`. In particular, optional parsers should support `std::optional` attributes.

saki7:
Part of #809

`x3::traits::move_to`: Dump tag dispatching and use concepts for overload resolution. This drastically improves compilation speed and prints significantly concise errors on ill-formed programs, thanks to the reduced nesting level on the entire control flow of X3. Also partially (but not yet completely) improves #346 due to reduced MPL usage.

`x3::detail::parse_into_container`: Ditto.

`support/traits/optional_traits.hpp`:
  `BOOST_SPIRIT_X3_USE_BOOST_OPTIONAL`: new macro.
  Default to `boost::optional` unless above macro is defined as `0`. Implements
  `std::optional` handling regardless of the value; fixes #270.
  Note that most test suites are intentionally not defining above macro as `1` 
  (yet) to make sure the change does not break existing codes.

`x3::optional`: Ideally this should've split into a separate PR, but the new attribute category handling requires the overhaul on this.

`core/proxy.hpp`: Removed. `x3::optional` was the only derived parser which used this feature for years. We shouldn't support any overly generic 'core' features whose extandable use case is unknown even to the core components.

`extract_int`, etc.: Adjusted to properly "move" the local attribute variable before passing it to `x3::traits::move_to`.

`char_parser`: Ditto, but this constructs temporary value instead of applying `std::as_const`. Const references are costly for primitive type like `Char` thus prvalue should be constructed here.

tests: No semantic changes intended. All changes exists solely for adapting to the new attribute category handling (implementation details.)


saki7:
Self-merged; these are implementation details and does not change the semantics of public API.

Issue #678, opened 2021-05-20
Since boost 1.72 spirit::unicode::char_ fails to parse non-ASCII


timo-schluessler:
Sample code to reproduce the issue:

```c++
#define BOOST_SPIRIT_UNICODE
#include <boost/spirit/include/qi.hpp>

int main()
{
   typedef std::string::const_iterator iterator_type;
   namespace qi = boost::spirit::qi;
   namespace unicode = boost::spirit::unicode;

   std::string input("\"Test ⏳\"");
   qi::rule<iterator_type, std::string(), unicode::space_type> quoted_string = qi::lexeme['"' >> +(unicode::char_ - '"') >> '"'];

   iterator_type iter = input.begin();
   iterator_type end = input.end();
   std::string output;
   bool r = phrase_parse(iter, end, quoted_string, unicode::space, output);

   if (r && iter == end)
      std::cout << "successfully parsed " << input << " to " << output << std::endl;
   else
      std::cout << "failed to parse " << input << std::endl;

   return 0;
}
```

Thanks to [sehe](https://stackoverflow.com/a/67592419/121331) who bisected the issue down to commit 16159fb.
Maybe this behavior is by intention - then I simply don't get the use and meaning of spirit::unicode and BOOST_SPIRIT_UNICODE.

Kojoley:
1. The code contains implementation-defined behavior ([\[lex/1.1\]](https://eel.is/c++draft/lex#phases-1.1)). Please rewrite it without unicode characters in the source code.
2. Do you intentionally use `std::string` with unicode? That will encode it with the execution character set.
3. '"' is not a unicode parser, probably it is the root cause of your issue.

timo-schluessler:
1. Sorry I didn't know this was implementation-defined. In the real program the string is read in from a file which is encoded in UTF-8. Please find the updated example below.
2. Yes. I would like to process the UTF-8 as plain 8-bit chars as the file format itself (say the control characters) is exclusively made up of ASCII characters, like the " in the example. The fields/values in the format though may contain any UTF-8 character. (And because of they way UTF-8 encodes unicode characters no single byte could be falsely interpreted as a valid ASCII control char.)
3. Do you have an idea how to fix this?

```c++
#define BOOST_SPIRIT_UNICODE
#include <boost/spirit/include/qi.hpp>

int main()
{
   typedef std::string::const_iterator iterator_type;
   namespace qi = boost::spirit::qi;
   namespace unicode = boost::spirit::unicode;

[...]

PR #817, opened 2025-09-08
[X3] Modernize char/string/symbols, make X3 totally self-contained

### LINKED ISSUE #798: char_encoding functions are not constexpr

saki7:
Part of #809
Fixes #798 

The diff is very large because I copied the `home/support/char_encoding/` components to X3's subdirectory. 
Modernization changes are about 2,000 additions and 1,800 deletions only.

## Summary

- `char_encoding` and `char_set` related components are copied from the `spirit/home/support/` directory to the `spirit/home/x3/` directory, and then modernized.
  - This would make X3 self-contained (finally!!!!!) and safely allows upgrading the codebase to C++23-style.
  - Note that other spirit components only support C++11, therefore adding C++23 support for X3 is impossible without this change. Macros are useless because there's no guarantee that the modernization can be expressed only by macros, and ODR violations would be engaged.
  - Unicode tables are copied as-is, with minor modernization patches.
- `x3::lit` and `x3::string` now uses concepts for overload resolution; provides significantly improved compilation speed and better errors.
- Mixture of incompatible character types (e.g. `char` and `wchar_t`) is now officially banned. Such wrong usage would invoke the deleted functions, safely alerting the end users. This would prevent issues like https://github.com/boostorg/spirit/issues/678#issuecomment-2376313379, which involves incorrect understanding of character encodings. Such misuse would introduce serious security issues and should be immediately banned.
- Add `constexpr` to almost everywhere.
- Rename `x3::symbols` to `x3::shared_symbols`.
  - Bonus: the new implementation has full allocator support, including `std::allocator_traits<allocator_type>::propagate_on_container_move_assignment`, etc.
  - Make `x3::symbols` the alias of `x3::shared_symbols`, with `[[deprecated]]` warning.
- `x3::unique_symbols`: Added; this is the `std::unique_ptr` counterpart of `x3::shared_symbols`.

## Deprecation of `ascii` and `iso8859_1`

Hard-coding ASCII or Latin-1 assumptions is almost always wrong in modern applications and **can introduce security vulnerabilities.**

Unless the input is provably ASCII-only (e.g., fixed-format data from the 1980s), the only correct interpretation uses the data's *original* encoding. Users should prefer transcoding to Unicode or to the implementation-defined execution character set before passing text to general-purpose parsers. For more information, refer to authoritative sources on C++ character encodings and Unicode fundamentals. 

## We should recognize the fact that the global `spirit/home/support/` directory had been a huge contribution barrier.

As a CJK (Japanese) speaker, I always wanted to contribute to the unicode components of X3. I even had tons of `constexpr` support even before this PR. However, **I hesitated to contribute for at least 5 years.** Why? Because the `char_encoding` functionalities were inside the global `spirit/home/support/` directory. 

That made me fear that **I would break four Boost Libraries**, because a single mistake in the `support/` directory would result in total breakage.

**The destination folder I chose is `spirit/home/x3/char_encoding/`, NOT `spirit/home/x3/support/char_encoding/`.** The subdirectory "support" is very poorly named, and even I always become confused by thinking what components are actually inside "support" and what are not. To be honest, EVERY "supportive" components can virtually fit inside the "support" subdirectory. We should stop adding new features to X3's "support" directory right now.

`char_set` components are NOT copied to the `x3/char_set/` directory. All components are actually implementation-detail char-related ones, which naturally fits inside the `x3/char/detail/` directory.


djowel:
> @djowel I'd like to hear your opinion, do you have any objections on the decisions I noted above?

I read through this the other day. I have no objections. Making X3 self-contained is a good goal.

saki7:
> I read through this the other day. I have no objections. Making X3 self-contained is a good goal.

Thanks for your understanding! I'll do my best to make the world better.

PR #819, opened 2025-09-09
Modernize `x3::confix`

saki7:
Part of #809 

Use concepts to constrain `x3::confix`. It is now a CPO that inhibits ADL.


saki7:
Self-merging; the changes are trivial and does not change the semantics of the public API.

PR #820, opened 2025-09-09
Modernize `x3::expect`

saki7:
Part of #809 
Blocked by #817 

Use concepts in `x3::expect`. It is also a CPO that inhibits ADL.


saki7:
Self-merged; this shouldn't introduce semantic changes on the public API.

PR #822, opened 2025-09-10
Modernize `x3::matches`

saki7:
Part of #809 

Use concepts in `x3::matches`. It is now a CPO that inhibits ADL.


saki7:
Self-merged; these are trivial and does not change the semantics of the public API.

PR #826, opened 2025-09-10
Modernize `x3::repeat`

saki7:
Part of #809 

- Use concepts in `x3::repeat`.
- `x3::repeat` is now a CPO that inhibits ADL.
- `x3::inf` is deprecated regarding overly generic name; use `x3::repeat_inf` instead.
- `x3::repeat[p]` is deprecated since it has the exact same meaning as `*p`. It is generally discouraged to provide multiple ways to achieve same the functionality in modern C++ library design.


saki7:
Self-merged. Although I added some deprecations, they are backward compatible as of now.

PR #830, opened 2025-09-11
Modernize numeric

saki7:
Part of boostorg/spirit_x4#4 

This PR modernizes `numeric` and `numeric_utils` components.

- Add `constexpr` to almost everywhere
- Use `if constexpr` in static branches
- Use iterator/sentinel pair, avoid unconstrained `typename Iterator`
- Add `noexcept` to primitive operations
- Use `<cstdint>` instead of Boost version
- Use `<type_traits>` instead of Boost version


saki7:
Superseded by boostorg/spirit_x4#17

PR #843, opened 2025-12-20
Karma: fix broken `real_generator` output for `long double`

### LINKED ISSUE #842: Karma: `real_generator` output for `long double` broken in Boost 1.90.0
After updating to Boost 1.90.0 my unit tests started failing. Looking into the problem revealed that Karma now generates incorrect output for certain `long double` values.

The following MWE demonstrates the problem:
```cpp
#include <iostream>
#include <iterator>
#include <limits>

#include <boost/spirit/home/karma/generate.hpp>
#include <boost/spirit/home/karma/numeric/real.hpp>
#include <boost/spirit/home/karma/numeric/real_policies.hpp>


template<class T>
struct max_precision_policy : boost::spirit::karma::real_policies<T> {
    static constexpr unsigned precision(T /*unused*/) noexcept {
        return std::numeric_limits<T>::max_digits10;
    }
};

int main()
{
    boost::spirit::karma::real_generator<long double,
                                         max_precision_policy<long double>>
        real;

    constexpr long double value = 1.55;
    boost::spirit::karma::generate(std::ostreambuf_iterator{std::cout}, real,
                                   value);
    std::cout << std::endl;
}
```
outputs the floating point value `1.55` as `1.0550000000000000044406` which is clearly incorrect. In Boost 1.89 the generated output is `1.550000000000000044406` as expected. The issue, however, is not limited to this specific value.

Bisecting points to e6fbdf615b43fa7a03db4877079e572481a55f35 as the bad commit.

sergiud:
Eliminate Boost 1.90.0 regression introduced in e6fbdf615b43fa7a03db4877079e572481a55f35.

Fixes #842

### DIFF
--- a/include/boost/spirit/home/karma/numeric/real_policies.hpp
+++ b/include/boost/spirit/home/karma/numeric/real_policies.hpp
@@ -257,20 +257,28 @@ namespace boost { namespace spirit { namespace karma
             //    generate(sink, right_align(precision, '0')[ulong], n);
             // but it's spelled out to avoid inter-modular dependencies.
 
-	    unsigned int digits=1; //should be number of digits n(truncating any fraction)
-	    if(!boost::spirit::traits::test_zero(n)) {
-	      static constexpr uint64_t limit = UINT64_MAX / 10;
-	      const T num = floor(n);
-	      for (uint64_t x = 10u, i = 1u;; x *= 10, i++) {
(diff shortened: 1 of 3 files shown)


hkaiser (approved):
LGTM, thanks!

saki7:
Thanks!

Issue #844, opened 2026-01-15
A redundant #include directive in move_to.hpp can be removed


ILer32:
Problem:
In [spirit/include/boost/spirit/home/x3/support/traits/move_to.hpp](https://github.com/boostorg/spirit/blob/develop/include/boost/spirit/home/x3/support/traits/move_to.hpp), 
one #include <boost/fusion/include/is_sequence.hpp> can be removed.
```
....
#include <boost/assert.hpp>
#include <boost/fusion/include/is_sequence.hpp> // 1
#include <boost/fusion/include/front.hpp>
#include <boost/fusion/include/move.hpp>
#include <boost/fusion/include/is_sequence.hpp> // 2
#include <utility>
```

djowel:
Will you kindly file a PR for this? Thanks!

ILer32:
The issue is solved.

PR #845, opened 2026-01-16
remove redundant include

ILer32:
The change is to address #844.


djowel:
Thank you!

ILer32:
My pleasure.

